PR #133, opened 2017-10-13
Feature/report/issue128 added class masters report

jrtaylor-com:


Hunteil:
Looks to be working. Also may want to update noun to be Paragon to match existing RoP (sorry, I know, it's annoying).

Hunteil:
Anyway to add a column for last attendance date? That way those aspiring for Paragon, can look at who is active.

PR #212, opened 2019-04-14
Issue 211 event date_start and add to player attendance

### LINKED ISSUE #211: Event date_start not honoured
If you run the following query the result should not include any events prior to 2018 but it does.  I believe the date string needs to be encased in quotes in the SQL query.  Will submit a pull request.

`https://amtgard.com/ork/orkservice/Json/index.php?&call=SearchService%2FEvent&kingdom_id=31&date_order=true&date_start=2018-01-01`

If you enclose the **$when** in single quotes (**'$when'**)then the query honours the date.

https://github.com/amtgard/ORK3/blob/master/system/lib/ork3/class.SearchService.php#L164

kenwalker:
Fixes #211 and also added new option for player attendance search for **date_start** so if you were looking for 6 months of attendance you could set date_start and not set limit.

### DIFF
--- a/system/lib/ork3/class.Player.php
+++ b/system/lib/ork3/class.Player.php
@@ -172,9 +172,13 @@ class Player extends Ork3 {
 						left join " . DB_PREFIX . "event e on a.event_id = e.event_id
 							left join " . DB_PREFIX . "park ep on e.park_id = ep.park_id
 							left join " . DB_PREFIX . "kingdom ek on e.kingdom_id = ek.kingdom_id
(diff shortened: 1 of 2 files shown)


kenwalker:
Have found an issue with this during testing.  Will update when I have a solution and what the issue was.

kenwalker:
I was quoting both the **date_start** AND default **$when** statement which did not need quoting.

kenwalker:
@mathus13 found an issue, I've pushed a change for it in latest commit

Issue #29, opened 2014-06-09
[Player] Waiver lost during creation


jrtaylor-com:
Old Ref: B73

Waiver information does not stick during player creation

PR #217, opened 2019-05-07
Waiver is always marked false

kenwalker:
Closes amtgard/ORK3#29 

Code checked if there was a waiver and an image and if not, set it to zero. But could be waivered AND no image.  In that case, set to 1.

Moved the code that set depending on **Waivered** being set to the last **else**


kenwalker:
Actually, could move that setter back up and just remove that last else case.  At least that what I dreamed last night 😢 ORK dreams now.  I'll fix this up tonight.

kenwalker:
Ok, fixed it up.

Issue #218, opened 2019-05-09
Cannot revoke dues or add dues via player update


kenwalker:
There's currently no way to revoke dues or add new dues in the Player API.

kenwalker:
Also noticed that Waivered was not honored and always set to false when using the API.

PR #219, opened 2019-05-09
Cannot revoke or add dues via player update

kenwalker:
Closes amtgard/ORK3#218

Add module load function, and load Treasury to revoke dues, returning the result

### DIFF
--- a/system/lib/ork3/class.Player.php
+++ b/system/lib/ork3/class.Player.php
@@ -605,10 +605,27 @@ class Player extends Ork3 {
 		}
 	}
 
+	public function load_model( $name )
+	{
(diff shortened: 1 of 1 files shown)


kenwalker:
This is required for my edit player page in my mobile application.

kenwalker:
I'm updating this PR to include adding dues. Will comment when done.

kenwalker:
Ok, I've added both in and the app currently works to revoke or add dues.

kenwalker:
Waivered in PlayerUpdate was checking for true, the _CreatePlayer_ path looks for 1 or 0.  Updated to honor that in player update

See **https://github.com/amtgard/ORK3/blob/master/system/lib/ork3/class.Player.php#L394**

PR #238, opened 2019-11-28
235 misplaced single quote in query

### LINKED ISSUE #235: Kingdom configuration for awards broken
Kingdom configuration is currently not allow for new or existing awards to be added. The drop down box is empty instead of containing all of the available awards and custom award type that should be there.

This was broken around the time that awards and positions were separated into different sections for the player view.

2019-09-10 16_26_14-ORK 3_ Editkingdom

kenwalker:
It looks like this fixes #235.  Maybe this worked before and new PHP? I'm not sure. Please can someone else pull and test before merging.
After change


kenwalker:
@esdraelon This I think should fix #235 (or does for my install)

PR #240, opened 2020-01-21
Add MundaneId as parameter to Guilds API

### LINKED ISSUE #239: Would like a MundaneId search for the Guilds report function
Currently the public _Guilds()_ function in Reports can take a KingdomId or a ParkId.  I'd like to add a MundaneId because I've had a request to show a player's eligible Guilds on the main player page within mORK without having to return the whole results for the players park and then filter on the client. This can save a lot of data going over the wire and mORK is a mobile app usually using cell data.

kenwalker:
Fixes #239


kenwalker:
So I can add the following to mORK 

image

Issue #255, opened 2020-07-03
Reverse order of heraldry & image


jkat718:
With the current order, the heraldry comes first and is therefore what is used in web previews (like on Facebook). Since, anecdotally, more people have images set than heraldry, this mean most ORK page profiles just have the placeholder image in their previews. By reversing the order, the more commonly-used field is on top and more accessible *and* would be used for previews.

jrtaylor-com:
This request doesn't actually require the template be "reordered". We just need to add an og meta tag in the header. EX: `<meta property="og:image" content="http://euro-travel-example.com/thumbnail.jpg">`
I dont think we are passing player info in the header at the moment, but I'll take a look

jrtaylor-com:
In the interim I've created the above PR that sets the ogimage meta tag to use the ORK's logo. This is a quick compromise for a better social media linking experience. At some point we can see about passing more data to the header that could be use logic to display either the headshot or the heraldry if they exist. At that point we could also update the title to be more descriptive which would help what your request is going for.

PR #258, opened 2021-04-12
Added ogimage meta tag that uses the ORK logo

jrtaylor-com:
Use the ORK logo rather than having social media automatically pull the first image which is likely an empty heraldry placeholder image.


kenwalker:
This seems ok for the use case. Overall I agree with the comments in #255 though.

PR #261, opened 2021-04-13
Issue237 resident date field

jrtaylor-com:
This PR requires a db-migration for the production DB. It is already in place on the DEV DB

### DIFF
--- a/system/lib/ork3/class.Player.php
+++ b/system/lib/ork3/class.Player.php
@@ -476,6 +476,7 @@ class Player extends Ork3 {
 				$this->mundane->active = $request['IsActive'];
 				$this->mundane->password_expires = date("Y-m-d H:i:s", time() + 60 * 60 * 24 * 365);
 				$this->mundane->password_salt = md5(rand().microtime());
+				$this->mundane->park_member_since = date('Y-m-d');
 				$this->mundane->save();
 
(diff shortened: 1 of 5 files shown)


system/lib/ork3/class.Player.php
```diff
@@ -215,8 +215,10 @@ public function GetPlayer($request) {
 					'HasImage' => $this->mundane->has_image,
 					'Image' => HTTP_PLAYER_IMAGE . sprintf('%06d.jpg', $this->mundane->mundane_id),
 					'PenaltyBox' => $this->mundane->penalty_box,
-          'Active' => $this->mundane->active,
-          'PasswordExpires' => $this->mundane->password_expires
+					'Active' => $this->mundane->active,
+					'PasswordExpires' => $this->mundane->password_expires,
+					//'ParkMemberSince' => date('d/m/Y', strtotime($this->mundane->park_member_since))
```
kenwalker, line 220: Can remove the commented out code?

kenwalker:
I tested this out locally, I also updated my local JSORK code to be able to show and update the value and I tried bogus values which did not fail but resulted in a date of "0000-00-00".  I tried transfers and they correctly set the date.
I tried a new player and it did not correctly set the date so I pushed a change which should fix this. Please have a look at that.

kenwalker:
I'm a bit worried that this field is blank for all players. I'm not sure of the right date we could set for it, certainly any player that took credits anywhere else? The first credit they took at their current park?

PR #262, opened 2021-04-13
Issue253 dropdown missing custom awards

jrtaylor-com:
Updated SQL and moved OfficerRole check to return array logic so that kingdom custom awards could be fetched. Fixed Award/Title bug stacking in Player award view


kenwalker:
While the "Order of the Roach" now shows up in the Dessert Winds dropdown, I can't seem to add it to a player as an award. I'm booted back to the login page. I'm using my Ork Admin account so should not be a privilege issue?

PR #263, opened 2021-04-13
Issue233 player image browser cache bust

jrtaylor-com:
Added a browser cache bust string that uses the record modified date for the player.


kenwalker:
This seems good

PR #264, opened 2021-04-13
Issue190 update failed login message

jrtaylor-com:


kenwalker:
If you try to login with an incorrect password with a username where the password has NOT expired you still get the same message. I'm not sure this current solution is any better?

PR #265, opened 2021-04-13
Issue192 rename player notes

jrtaylor-com:


kenwalker:
This looks good, extra changes with the ork site rename? Those look good too.

PR #266, opened 2021-04-13
Issue115 reeve corpora fields and reports

jrtaylor-com:
This PR requires a DB migration. Already done for testing on DEV DB


kenwalker:
This looks good too but will have to be updated after #261 is committed since these both have changes to the player UI. I also hacked my own jsork tools to manipulate these fields and it worked fine. 
If you set the dates in the past, the reports do the right thing and not show the players, but in the player UI it's not really clear if they're qualified or not. Perhaps the date field could be red if the date is in the past?

PR #268, opened 2021-04-18
Added file size info to admin image box, added file size errors

jrtaylor-com:
This file size takes into account the added 1.333 bytes per byte that the base64 encode that is used on the backend adds

### DIFF
--- a/orkui/controller/controller.Admin.php
+++ b/orkui/controller/controller.Admin.php
@@ -682,15 +682,20 @@ class Controller_Admin extends Controller {
 					    }
 						if ($this->request->Update == 'Update Media') {
 							if ($_FILES['Heraldry']['size'] > 0 && Common::supported_mime_types($_FILES['Heraldry']['type'])) {
-								if (move_uploaded_file($_FILES['Heraldry']['tmp_name'], DIR_TMP . sprintf("h_%06d", $id))) {
-									$h_im = file_get_contents(DIR_TMP . sprintf("h_%06d", $id));
-									$h_imdata = base64_encode($h_im);
-									$this->Player->SetHeraldry(array(
-										'MundaneId' => $id,
-										'Heraldry' => strlen($h_imdata)>0?$h_imdata:null,
-										'HeraldryMimeType' => strlen($h_imdata)>0?$_FILES['Heraldry']['type']:'',
-										'Token' => $this->session->token
-									));
+								if ((int) $_FILES['Heraldry']['size'] * 1.333 > 465000) {
+									$this->data['Error'] = 'Image Error: File size is to large.';
+									$r['Status'] = NULL;
+								} else {
+									if (move_uploaded_file($_FILES['Heraldry']['tmp_name'], DIR_TMP . sprintf("h_%06d", $id))) {
+										$h_im = file_get_contents(DIR_TMP . sprintf("h_%06d", $id));
+										$h_imdata = base64_encode($h_im);
+										$this->Player->SetHeraldry(array(
+											'MundaneId' => $id,
+											'Heraldry' => strlen($h_imdata)>0?$h_imdata:null,
+											'HeraldryMimeType' => strlen($h_imdata)>0?$_FILES['Heraldry']['type']:'',
+											'Token' => $this->session->token
+										));
+									}
 								}
 							}
 							if ($_FILES['Waiver']['size'] > 0 && Common::supported_mime_types($_FILES['Waiver']['type'])) {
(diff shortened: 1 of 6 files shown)


orkui/controller/controller.Admin.php
```diff
@@ -682,15 +682,20 @@ public function player($id) {
...
-									$h_im = file_get_contents(DIR_TMP . sprintf("h_%06d", $id));
-									$h_imdata = base64_encode($h_im);
-									$this->Player->SetHeraldry(array(
-										'MundaneId' => $id,
-										'Heraldry' => strlen($h_imdata)>0?$h_imdata:null,
-										'HeraldryMimeType' => strlen($h_imdata)>0?$_FILES['Heraldry']['type']:'',
-										'Token' => $this->session->token
-									));
+								if ((int) $_FILES['Heraldry']['size'] * 1.333 > 465000) {
+									$this->data['Error'] = 'Image Error: File size is to large.';
```
kenwalker, line 686: "too large" same with one below

orkui/template/default/Admin_player.tpl
```diff
@@ -25,6 +25,14 @@
 			<span></span>
 			<span><input type='submit' value='Update Media' name='Update' /></span>
 		</div>
+		<div>
+			<span></span>
+			<span style="font-size: 10px;">Supported Types: JPG, GIF, PNG</span>
+		</div>
+		<div>
+			<span></span>
+			<span style="font-size: 10px;">Max Size: 2725 Kb (348836 bytes)</span>
```
kenwalker, line 34: Most people are not familiar with Kilobits and most file information within OSs is shown in KB or MB. I suggest changing to KB and omitting bytes all together.

kenwalker:
Have left some comments above but otherwise the functionality works. It's a bit weird that you get both a green "Player updated" and the "Image error".  Not a big deal though.
image

jrtaylor-com:
@kenwalker I added the spelling mods, and modified the image size text. I kept Bytes since both OSs show bytes when inspecting a file. Converted kilobits to kilobytes.

Screen Shot 2021-04-18 at 9 20 23 AM
Screen Shot 2021-04-18 at 9 19 39 AM

jrtaylor-com:
@kenwalker added officers to kingdom/park index page 
Screen Shot 2021-04-18 at 10 21 51 AM
Screen Shot 2021-04-18 at 10 21 41 AM

kenwalker:
Monarchy info on both pages looks good

kenwalker:
There are approximately 39 parks (according to the sql query) that have not put http:// or https:// in front of their URL and in that case your link does not work properly. For example, try Felfrost's link.

jrtaylor-com:
> There are approximately 39 parks (according to the sql query) that have not put http:// or https:// in front of their URL and in that case your link does not work properly. For example, try Felfrost's link.

I feel like once they see the feature they might fix their links. They didn't have a reason to before.

kenwalker:
> > There are approximately 39 parks (according to the sql query) that have not put http:// or https:// in front of their URL and in that case your link does not work properly. For example, try Felfrost's link.
> 
> I feel like once they see the feature they might fix their links. They didn't have a reason to before.

I volunteer as tribute to just go fix them ha.  I wonder if we could add a validation to the setting of it for any future entries to ensure it has http:// or https:// ?

jrtaylor-com:
> > > There are approximately 39 parks (according to the sql query) that have not put http:// or https:// in front of their URL and in that case your link does not work properly. For example, try Felfrost's link.
> > 
> > 
> > I feel like once they see the feature they might fix their links. They didn't have a reason to before.
> 
> I volunteer as tribute to just go fix them ha. I wonder if we could add a validation to the setting of it for any future entries to ensure it has http:// or https:// ?

I expect a lot of them to be outdated links anyhow. the http/https would be dependent on the host themselves, not on the ORK server. You could replace them with `//` instead of `https:` or `http:` which would just use the ORK's current state. This would keep them from throwing a mixed content error when the ORK is using https.

Only issue would be if the target server doesn't handle switching between http/https gracefully

kenwalker:
> > > > There are approximately 39 parks (according to the sql query) that have not put http:// or https:// in front of their URL and in that case your link does not work properly. For example, try Felfrost's link.
> > > 
> > > 
> > > I feel like once they see the feature they might fix their links. They didn't have a reason to before.
> > 
> > 
> > I volunteer as tribute to just go fix them ha. I wonder if we could add a validation to the setting of it for any future entries to ensure it has http:// or https:// ?
> 
> I expect a lot of them to be outdated links anyhow. the http/https would be dependent on the host themselves, not on the ORK server. You could replace them with `//` instead of `https:` or `http:` which would just use the ORK's current state. This would keep them from throwing a mixed content error when the ORK is using https.
> 
> Only issue would be if the target server doesn't handle switching between http/https gracefully

I mean I’ll try them out individually and fix their info.

PR #273, opened 2021-04-20
Updated Report class GetActivePlayers method to use new dues table

jrtaylor-com:


jrtaylor-com:
Testing: Ensure reports display correct information
Routes:
`?Route=Reports/active_duespaid/Kingdom&id=14`
`?Route=Reports/active_waivered_duespaid/Kingdom&id=14`

kenwalker:
This looks good. I did find a weird dues paid player entry in that it says the Kingdom is Goldenvale but the Park is Wolvenfang. So they show up as dues paid in Wolvenfang but not in the active list because they're from a different park. And they don't show up as active dues paid in that park because the dues are in Goldenvale for some reason. Perhaps revoking those dues and adding legit one would solve it but there might be others?

https://ork.amtgard.com/orkui/index.php?Route=Player/index/66590

kenwalker (approved):

PR #276, opened 2021-04-23
URL field added to edit unit page

tzulah:
I wasn't able to test whether URL changes worked or not, but the field is there now, and it does populate itself with existing URL entry if there is one.


jrtaylor-com (approved):
Looks good

PR #280, opened 2021-05-04
Modified pw reset email for better yahoo parsing

jrtaylor-com:


kenwalker (approved):
I think this looks good, having issues trying it locally with my dev server/email

PR #281, opened 2021-05-06
2021 may 03 release

jrtaylor-com:


kenwalker:
After the last fix I was able to successfully use Yahoo, Outlook, Gmail, my random email server, web interfaces and MacOS and iOS mail clients

PR #286, opened 2021-05-19
Fix array_merge error on knights and masters report

jrtaylor-com:


kenwalker (approved):
Looks good, test passed

PR #296, opened 2021-08-08
SQL is pulling mundane_id from authorization db which Champion is not in

### LINKED ISSUE #295: Chapter/Kingdom set officers page has broken link for champion
Link to user account that is set as champion does not include player id. Occurs on chapter and kingdom level admin pages.

Example here: https://ork.amtgard.com/orkui/index.php?Route=Admin/setkingdomofficers&KingdomId=1

kenwalker:
Fixes #295 The mundane_id was being pulled from the join with the authorization DB which the Champion is not in so many fields were set to null including mundane_id. Set an additional field in the sql results to be the mundane_id and used that in the results. 

@jrtaylor-com you can make a branch for a release soon and change the merge destination if you want.

image

image

### DIFF
--- a/system/lib/ork3/class.Kingdom.php
+++ b/system/lib/ork3/class.Kingdom.php
@@ -495,7 +495,7 @@ class Kingdom  extends Ork3 {
 	}
 
 	public function GetOfficers($request) {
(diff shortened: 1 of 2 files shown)


jrtaylor-com (approved):
Looks good, I updated to merge into the release branch 2021AUG-09-RELEASE.

PR #306, opened 2021-09-06
Remove Default Awards for Parks/Kingdoms

jkat718:
The Add Award dialogue for Parks and Kingdoms still had "Archduchess" as the default, whereas the Add Award dialogue for Players defaults to blank (to avoid accidentally awarding the first award alphabetically). This should fix that issue, by adding a blank entry to the top of the list, the same way the Player dialogue was changed.


jrtaylor-com:
Thanks! Will push it out with a release later this week

PR #311, opened 2022-01-25
Enhancement to return entered_at to attendance

### LINKED ISSUE #310: Attendance queries do not return when they were entered
A player requested that they be able to find out when an attendance was entered, and that value is in the database as **entered_at** but is not returned by the API calls to the ORK.

kenwalker:
Fixes #310 

Added **entered_at** to attendance call results. Also fixed a typo which was preventing **Attendance_Id** from being returned in player attendance record. 

**Park getAttendance now**
Park getAttendance

**Event getAttendance now**
Event getAttendance

**Player getAttendance now**
Player getAttendance

### DIFF
--- a/system/lib/ork3/class.Player.php
+++ b/system/lib/ork3/class.Player.php
@@ -282,7 +282,9 @@ class Player extends Ork3 {
 		} else if ($r->size() > 0) {
 			do {
 				$response['Attendance'][] = array(
-						'AttendanceId' => $r->attenance_id,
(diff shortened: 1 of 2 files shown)


jrtaylor-com:
@kenwalker I missed this but will get it into the next release later this week or next

jrtaylor-com (approved):

kenwalker:
Cheers @jrtaylor-com

PR #330, opened 2023-02-07
Update class.Player.php to actually fetch companies for players

megiddo:


kenwalker:
Updated return value to be the Units array and removed the "Company" div in the player admin page since it doesn't make sense.
Even though there is kind of an unwritten rule of only belonging to 1 company there are over 1500 players who have more than 1 and there's 150 with 4 or more. So it might be hard to enforce 1 company at this point.
This shouldn't impact any 'jsork' or 'mORK' or any other API user because the 'Company' field in the result of the player info was blank forever.

Issue #345, opened 2024-09-02
Unnecessary exposure of names in some pages and reports


kenwalker:
As originally designed some extra information about officers was included in certain reports. This should be changed to limit exposure

kenwalker:
This was fixed and deployed

PR #347, opened 2024-09-04
Fix for Unnecessary exposure of names #345

kenwalker:
The API call to GetOfficers has been modified to respect the authorization level of the token being used to invoke the call. So a park officer with edit rights may see the additional information in the edit officers page or if they pass a valid token to the API they will get the additional information, same for kingdom, same for admin.

### DIFF
--- a/orkui/model/model.Login.php
+++ b/orkui/model/model.Login.php
@@ -12,6 +12,11 @@ class Model_Login extends Model {
 		unset($this->session->user_name);
 		unset($this->session->token);
 		unset($this->session->timeout);
+		unset($this->session->kingdom_id);
+		unset($this->session->kingdom_name);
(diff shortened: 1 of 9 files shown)


orkui/model/model.Login.php
```diff
@@ -12,6 +12,11 @@ function logout($userid) {
 		unset($this->session->user_name);
 		unset($this->session->token);
 		unset($this->session->timeout);
+		unset($this->session->kingdom_id);
+		unset($this->session->kingdom_name);
+		if (isset($_SESSION['is_authorized_mundane_id'])) {
```
kenwalker, line 17: Even after logout this key was still around and it is used in the authorization code so I'm explicitly clearing it on logout

Issue #206, opened 2018-12-23
Minus "-" in usernames breaks some searches


Glenalth:
Searching for usernames with a minus "-" does not always work. It also seems to be the default separator when the system is creating a username with an appended random string for uniqueness.

PR #363, opened 2025-06-08
Fixed player search to allow '-' and spaces. Plus support for full name search (like "Jane Doe")

prileo:
As per reported issue https://github.com/amtgard/ORK3/issues/206 '-' character break the player search as it is a dedicated operator in MATCH AGAINST search clause. However, many usernames include '-'

To keep the search fast and solve this issue: 
1) made MATCH AGAINST clause focus only on the first word entered into the search field. 
2) added a set of LIKE clauses to catch everything after that: with dasher, spaces, etc. 
3) included a concat(`given_name`,' ',`surname`) to support full name search.

### DIFF
--- a/system/lib/ork3/class.SearchService.php
+++ b/system/lib/ork3/class.SearchService.php
@@ -317,7 +317,6 @@ class SearchService extends Ork3 {
 					limit $limit";
 		$i = 0;
 		$this->db->clear();
-		error_log("olgafix: " . $sql, 0);
 		$q = $this->db->query($sql);
 		if ($q !== false && $q->size() > 0) {
 			$r = array();


kenwalker:
Seems to work much better. Verified that API calls work as well. Thanks for your contribution!!!

kenwalker (commented):
Looks good

esdraelon:
In the future, let's squash PRs into a single commit. Makes it a bit easier to review. Thanks!

system/lib/ork3/class.SearchService.php
```diff
@@ -289,7 +289,9 @@ public function Player($type, $search, $limit=15, $kingdom_id = null, $park_id =
 				break;
 			default:
 				$zztop = implode('* ', $searchtokens) . '*';
-				$s = "match(`given_name`, `surname`, `other_name`, `username`, `persona`) against ('" . mysql_real_escape_string($zztop) . "' in boolean mode)";
+				$s = "match(`given_name`, `surname`, `other_name`, `username`, `persona`) against ('" . mysql_real_escape_string($zztop) . "' in boolean mode)
+				and (`given_name` like '%" . mysql_real_escape_string($search) . "%' or `surname` like '%" . mysql_real_escape_string($search) . "%' or `username` like '%" . mysql_real_escape_string($search) . "%'
+				or `other_name` like '%" . mysql_real_escape_string($search) . "%' or `persona` like '%" . mysql_real_escape_string($search) . "%')";
```
esdraelon: Prefixed wildcard searches are un-indexable full table scans. We don't have great visibility on resource uses here, but we should watch to make sure there aren't degradations in return latency.

Issue #326, opened 2022-12-13
Make Park Titles a little more prominent in the UI


kenwalker:
Currently to see a parks title (Shire, Barony, etc) you have to do to the Kingdom Admin page and configure parks.

The titles could be added to the table of parks and as a new section in the Park overview?

PR #375, opened 2025-12-13
Park title prominence 326

ScottGrogin:
This PR addresses issue https://github.com/amtgard/ORK3/issues/326

## Changes
- The individual if statements used to display ParkInfo data was abstracted out into a loop over an associative array where the keys contain the actual text to be displayed on the page.
- Additionally .DS_Store and vs code config files were added to gitignore

Screenshot 2025-12-12 at 10 23 10 PM


kenwalker:
Something that is more noticeable with this change is that if a Park has no description, you see the Park title then the list of things you can do. 
image
image

Another tweak should be some padding between the title and the next section? 
image

Issue #327, opened 2023-01-24
Enter Award page does not account for custom award name dropdown


kenwalker:
If you are on a player's admin page and go to enter an award and choose "Custom Award", there is logic in Admin_player.tpl to show and hide a field for the 'CustomAwardName' and also use this information when saving the award.  

The Award_addawards.tpl used by the park and kingdom Add Awards option does not have this logic so you cannot name the award.

PR #379, opened 2025-12-24
Fix/327 custom award name

ScottGrogin:
This PR addresses https://github.com/amtgard/ORK3/issues/327.

I added code to the Award_addawards.tpl file to conditionally display an "Award Name" field for custom awards and added the field to the award controller so it actually shows up on a users page after being entered.

Additionally, I refactored the controllers code since the kingdom and park functions were essentially identical with the only difference between the 2 being what route they used. I made these changes over many commits to make them easier to follow.

Award entry on kingdom page:
Screenshot 2025-12-23 at 10 49 40 PM

Result:
Screenshot 2025-12-23 at 10 50 13 PM

Award entry on park page:
Screenshot 2025-12-23 at 10 47 12 PM

Result:
Screenshot 2025-12-23 at 10 48 17 PM

### DIFF
--- a/orkui/template/default/Admin_player.tpl
+++ b/orkui/template/default/Admin_player.tpl
@@ -594,7 +594,7 @@
 			<span>Award Name:</span>
 			<span><input type='text' value='<?=isset($Admin_player)?$Admin_player['AwardName']:$Player['AwardName'] ?>' name='AwardName' id='AwardName' /></span>
 		</div>
-		<div>
+		<div id='AwardRankField'>
 			<span>Rank:</span>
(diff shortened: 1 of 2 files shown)


kenwalker:
May I add a request in since you're hiding/showing fields? 

Add an ID to `<div id='AwardRankField'>` and hide/show it depending on if Officers is on or not. We don't need a Rank field for officers.

ScottGrogin:
> May I add a request in since you're hiding/showing fields?
> 
> Add an ID to `<div id='AwardRankField'>` and hide/show it depending on if Officers is on or not. We don't need a Rank field for officers.

I made sure to do all of this for the Admin_player file and the Award_addawards page

PR #384, opened 2026-02-08
Enhancement: Adding last sign-in date to the player profile for quick reference.

baltinerdist:
Currently, a user must scroll down to the attendance section to identify the last time a given player signed in. This adds last sign-in date as a common function and displays it in the Player Details box.

This also adds last sign-in date as a sorting mechanism for the Player Heraldry report.


esdraelon (changes requested):
Let's collapse this logic back without the branch and simply output the lastsignin if it exists or null otherwise.

Issue #305, opened 2021-09-06
[Feature Request] Sorting/Filtering Other Awards Tables


jkat718:
Similar to the changes Fog made to the Award Recommendation report, but for the tables of awards that have actually been given (on player profiles and/or on award-related reports). Useful for sorting by date rather than alphabetically, filtering on "Awarded By" or "Entered By" columns, etc.

PR #385, opened 2026-02-08
Feature/sortable-tables

baltinerdist:
Enables table sorting in the player Awards, Attendance, and Titles tables per the request in amtgard/ORK3#305


kenwalker:
Looks good thanks

PR #393, opened 2026-02-13
Enhancement: Prevent incomplete awards from being submitted

### LINKED ISSUE #380: When adding an award either in player, park, kingdom, the Add button should be disabled initially
In any of the places you can add an award, it has the entry fields empty, you can press the **Add** button and immediately you get an error. 

Look into disabling this button initially and only enabling it at a minimum when required fields are filled.

Also see if the logic for displaying both an error, but success can be aligned. There can only be one! Highlander rules.

Image

Additionally, it looks like the error Is not cleared and so shows even if the next Add is successful.

baltinerdist:
## Summary
- Disables the **Add** button on award forms until all required fields are filled (award selection, recipient, given-by)
- Fixes bug where both error and success messages could display simultaneously after submission ("Highlander" bug)
- Adds missing `break` in Admin controller validation preventing fall-through to API call on error

## Details
Addresses all three issues raised in #380:

**1. Add button disabled by default** — At kingdom/park level (`Award_addawards.tpl`) and admin player level (`Admin_player.tpl`), the submit button starts disabled and enables via `checkRequiredFields()` JS validation when the required fields have values.

**2. Error/success mutual exclusivity** — In `controller.Award.php`, moved result handling inside the validation-passed block so `handle_success()` is never called when validation fails. In `controller.Admin.php`, added a guard (`strlen($this->data['Error']) == 0`) around the result handler to prevent setting success when an error already exists.

**3. Missing break on GivenBy validation** — `controller.Admin.php` was missing a `break` after the "Who gave this award?" error, causing it to fall through and call `add_player_award()` anyway.

## Files Changed
- `orkui/controller/controller.Award.php` — Fix error/success mutual exclusivity
- `orkui/controller/controller.Admin.php` — Add missing `break`, guard dual messages
- `orkui/template/default/Award_addawards.tpl` — Disable button + JS validation
- `orkui/template/default/Admin_player.tpl` — Disable button + JS validation

## Test plan
- [ ] Navigate to kingdom add-award page — confirm Add button is disabled until Award, GivenTo, and GivenBy are selected
- [ ] Navigate to park add-award page — same verification
- [ ] Navigate to Admin > Player — confirm Add button disabled until Award and GivenBy are selected
- [ ] Submit with valid data — only success message appears
- [ ] Trigger a backend error — only error message appears (no success alongside)

Closes #380

### DIFF
--- a/orkui/controller/controller.Award.php
+++ b/orkui/controller/controller.Award.php
@@ -99,11 +99,12 @@ class Controller_Award extends Controller
 	}
 	private function add_award()
 	{
+		$isOfficer = $this->request->Award_addawards->awardtype === 'officers';
 		return $this->Player->add_player_award(array(
 			'Token' => $this->session->token,
 			'RecipientId' => $this->request->Award_addawards->MundaneId,
 			'KingdomAwardId' => $this->request->Award_addawards->AwardId,
-			'Rank' => $this->request->Award_addawards->Rank,
+			'Rank' => $isOfficer ? null : $this->request->Award_addawards->Rank,
 			'Date' => $this->request->Award_addawards->Date,
 			'GivenById' => $this->request->Award_addawards->GivenById,
 			'Note' => $this->request->Award_addawards->Note,
--- a/orkui/template/default/Award_addawards.tpl
+++ b/orkui/template/default/Award_addawards.tpl
@@ -27,6 +27,7 @@
 			if($(this).val() == 'officers'){
 				$( '#AwardNameField' ).hide();
 				$( '#AwardRankField' ).hide();
+				$( '#Rank' ).val('');
 			}else{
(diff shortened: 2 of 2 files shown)


baltinerdist:
Fixed in the latest commit by stripping the rank on two levels:

- **Frontend** (`Award_addawards.tpl`): When the user switches to the Officers radio button, the Rank field is now cleared (`$('#Rank').val('')`) in addition to being hidden. This prevents a previously-entered rank value from lingering in the hidden field and being submitted.

- **Backend** (`controller.Award.php`): `add_award()` now checks the submitted `awardtype` value and passes `null` for Rank when it's `'officers'`, so any rank data is stripped server-side as a second line of defense even if the UI is bypassed.

PR #402, opened 2026-02-15
Enhancement: Updates to image upload and view

baltinerdist:
## Summary

- **Auto-resize oversized images before upload** — When a user selects an image that exceeds the 340 KB limit, the browser uses the Canvas API to proportionally scale it down (up to 5 passes at JPEG quality 0.85) and replaces the file input's value via `DataTransfer` so the form submits the resized version transparently. An inline notice appears next to the file input showing the original and final size plus new dimensions (e.g. _Resized 850 KB → 298 KB (640×480)_). A generation counter prevents a race condition where a stale async resize callback could overwrite a replacement file the user selected while the first resize was still running.

- **Cache-bust heraldry thumbnails on admin edit pages** — All heraldry/image `` tags in the five admin edit templates (`Admin_player.tpl`, `Admin_editpark.tpl`, `Admin_editkingdom.tpl`, `Admin_unit.tpl`, `Admin_event.tpl`) now append `?t=<?=time()?>` to their `src`. This ensures the browser always fetches the latest file after an upload instead of serving a stale cached copy.

- **Click-to-expand lightbox for heraldry thumbnails** — Clicking any `.heraldry-img` thumbnail opens a full-viewport dark overlay showing the image at up to 90vw/90vh. Click the overlay or press Escape to dismiss. Images that are already wrapped in `<a>` links (e.g. the heraldry report, where clicking navigates to the player page) are excluded from the lightbox so their existing navigation behaviour is preserved.

## Files changed

| File | Change |
|---|---|
| `orkui/template/default/script/default.js` | Auto-resize helper + enhanced file input handler + lightbox JS |
| `orkui/template/default/script/orkui.js` | Same changes mirrored into the production bundle |
| `orkui/template/default/style/orkui.css` | Lightbox overlay + `cursor: pointer` on `.heraldry-img` |
| `Admin_player.tpl` | Cache-bust heraldry + player image src |
| `Admin_editpark.tpl` | Cache-bust park heraldry src |
| `Admin_editkingdom.tpl` | Cache-bust kingdom heraldry src |
| `Admin_unit.tpl` | Cache-bust unit heraldry src |
| `Admin_event.tpl` | Cache-bust event heraldry src |

## Test plan

- [ ] Select an image **under 340 KB** on an admin edit page → uploads normally, no resize notice shown
- [ ] Select an image **over 340 KB** → inline notice appears, form submits successfully with the resized image visible in the heraldry slot
- [ ] After selecting a large image (resize in flight), quickly replace with a small image → small image is what actually uploads (generation counter prevents the stale resize from overwriting it)
- [ ] After uploading a new heraldry image, reload the admin edit page → new image is shown immediately (no stale cache)
- [ ] Click a heraldry thumbnail on any non-report page → lightbox overlay opens showing the full image; click overlay or press Escape to dismiss
- [ ] Click a heraldry thumbnail in the heraldry report → navigates to player page as before (lightbox does not intercept)

### DIFF
--- a/orkui/template/default/Admin_editkingdom.tpl
+++ b/orkui/template/default/Admin_editkingdom.tpl
@@ -7,7 +7,7 @@
 		<div>
 			<span>Heraldry:</span>
 			<span>
-				<img class='heraldry-img' src='<?=$Kingdom_heraldryurl['Url'] ?>' />
+				<img class='heraldry-img' src='<?=$Kingdom_heraldryurl['Url'] . '?t=' . time() ?>' />
 				<input type='file' name='Heraldry' id='Heraldry' />
 			</span>
 		</div>
--- a/orkui/template/default/Admin_editpark.tpl
+++ b/orkui/template/default/Admin_editpark.tpl
@@ -11,7 +11,7 @@
 		<div>
 			<span>Heraldry:</span>
 			<span>
-				<img class='heraldry-img' src='<?=$Park_data['HasHeraldry']?$Park_heraldry['Url']:HTTP_PLAYER_HERALDRY . '000000.jpg' ?>' />
+				<img class='heraldry-img' src='<?=($Park_data['HasHeraldry']?$Park_heraldry['Url']:HTTP_PLAYER_HERALDRY . '000000.jpg') . '?t=' . time() ?>' />
 				<input type='file' name='Heraldry' class='restricted-image-type' id='Heraldry' />
 			</span>
 		</div>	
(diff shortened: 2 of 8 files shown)


kenwalker:
This doesn't appear to work on Safari. I see the resizing on Chrome.
The click-to-expand also does not work on Safari. I see it on Chrome.

kenwalker:
Downloaded and tried Firefox and it's fine there, including the click-to-expand

PR #421, opened 2026-02-23
Fix double-counting in Park Attendance Explorer report

### LINKED ISSUE #1: v8 Level Options
Option in Park config to display v8 Levels.

[Tuncer] I think having the v7 and v8 levels appear as a checkbox in the park config would be good for this. Depending on whether or not the option is selected that level column will appear. At least one option must be selected and until v8 is rolled out v7 should be defaulted. There should be a 'catch all' in the display code to show v7 if no option is selected.

### LINKED ISSUE #2: JSON-based Federation
The existing federation system uses SOAP.

I have a working federation model (from another project), that leverages JSON-based RPCs, and definitely works.  Some care will have to be taken around authentication and session storage, but it will probably be easier than making sure that every last SOAP call is properly constructed in the WSDL, especially if we plan on having a moving target for the API for a while.

baltinerdist:
## Summary

Two bugs in the Park Attendance Explorer caused attendance figures to be silently overcounted.

### Bug 1 — Dues JOIN multiplied `signin_count` per player (Individual Park view)

`ParkAttendanceSinglePark` joins `ork_dues` with a `LEFT JOIN` that matches all active, non-revoked dues rows. Players with more than one valid dues record (e.g. a lifetime dues row **and** an unexpired regular dues row, or two overlapping payment windows) caused a Cartesian product with each of their attendance rows. `COUNT(*)` then counted the multiplied rows rather than the actual sign-ins.

**Fix:** `COUNT(*)` → `COUNT(DISTINCT a.attendance_id)` so the aggregate counts attendance records, not join output rows. The `MAX()` aggregates used for dues display are unaffected.

### Bug 2 — Kingdom Totals double-counted cross-park visitors (All Parks view)

The Kingdom Totals summary row was built by summing `UniquePlayers` and `UniqueMembers` from each per-park row. A player who signed in at multiple parks was counted once per park, inflating both figures.

**Fix:** A new SQL query is added to `ParkAttendanceAllParks()` that computes `COUNT(DISTINCT a.mundane_id)` and `COUNT(DISTINCT CASE WHEN m.park_id = a.park_id THEN a.mundane_id END)` in a single pass across the whole kingdom. These true distinct counts flow through the model and controller into the summary row. `TotalSignins` and `Members 2+/3+/4+` continue to be summed from per-park rows, which is correct for those fields.

The "About This Report" copy is updated to reflect the corrected behaviour.

## Files changed

- `system/lib/ork3/class.Report.php` — Fix #1 (`COUNT(DISTINCT)`); add summary query for Fix #2
- `orkui/model/model.Reports.php` — Surface `Summary` key alongside `Attendance`
- `orkui/controller/controller.Reports.php` — Use summary for `UniquePlayers`/`UniqueMembers` in kingdom totals
- `orkui/template/default/Reports_parkattendanceexplorer.tpl` — Update "About This Report" note

## Test plan

- [ ] Run All Parks report for a kingdom; confirm Kingdom Totals Unique Players/Members are ≤ the sum of per-park values (they should be equal only when no player visited more than one park)
- [ ] Run Individual Park report for a player known to have multiple active dues records; confirm per-period sign-in counts match actual attendance, not a multiple thereof
- [ ] Verify Dues Paid column still shows correct value (Life / expiry date / blank)
- [ ] Spot-check Total Sign-Ins and Members 2+/3+/4+ are unchanged from previous behaviour

### DIFF
--- a/orkui/controller/controller.Reports.php
+++ b/orkui/controller/controller.Reports.php
@@ -494,24 +494,25 @@ class Controller_Reports extends Controller {
 			$this->data['mode'] = 'all_parks';
 
 			if (is_array($result)) {
-				$this->data['attendance'] = $result;
+				$attendance = $result['Attendance'];
+				$kingdom_summary = $result['Summary'];
+				$this->data['attendance'] = $attendance;
 
-				// Compute kingdom summary
+				// Compute kingdom summary row; UniquePlayers/UniqueMembers come from the
+				// server-side distinct query to avoid double-counting cross-park visitors.
 				$summary = array(
 					'TotalSignins' => 0,
-					'UniquePlayers' => 0,
-					'UniqueMembers' => 0,
+					'UniquePlayers' => (int)($kingdom_summary['UniquePlayers'] ?? 0),
+					'UniqueMembers' => (int)($kingdom_summary['UniqueMembers'] ?? 0),
 					'Members2Plus' => 0,
 					'Members3Plus' => 0,
 					'Members4Plus' => 0,
 					'WeeksInPeriod' => 0,
 					'MonthsInPeriod' => 0,
-					'RowCount' => count($result)
+					'RowCount' => count($attendance)
 				);
-				foreach ($result as $row) {
+				foreach ($attendance as $row) {
 					$summary['TotalSignins'] += $row['TotalSignins'];
-					$summary['UniquePlayers'] += $row['UniquePlayers'];
-					$summary['UniqueMembers'] += $row['UniqueMembers'];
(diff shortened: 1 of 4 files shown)


baltinerdist:
@kenwalker  I've ran the numbers on the redacted database you gave me on the bugfix/PAE-report-doubles branch and the numbers are good. Can you double check to make sure you've got that branch loaded in?

kenwalker:
@baltinerdist seems fine now. Maybe I was testing on wrong branch.

Issue #422, opened 2026-02-23
Allow special characters in player fields


kenwalker:
Currently you cannot use characters like é while creating a character. There are also many characters with fields with **`\'`** sequence of characters. These should work and should be replaced throughout the database with **`'`**

PR #423, opened 2026-02-24
Change regex and allow  UTF-8

kenwalker:
Fix for #422 

There are also sql changes in the issue that have to be run.

image
image


megiddo:
YOLO on a local instance and see what happens.

PR #431, opened 2026-03-17
Experiment: Voronoi-based kingdom territory map

### LINKED ISSUE #151: Automatically Generate Kingdom Borders
Use voronoi or a similar method to automatically generate kingdom boundaries based on park locations (and weight, if possible).

baltinerdist:
Closes #151

## Summary

Adds `map/build_map.py` — a standalone Python script that generates an interactive Leaflet/OpenStreetMap HTML visualization of Amtgard kingdom territories.

- **Voronoi diagram** — each active, non-freehold park is a node; every point on the map belongs to the nearest park's kingdom
- **Connected territories** — adjacent same-kingdom parks form one contiguous region with no gaps
- **Land-clipped** — territories are constrained to US and Canada borders (Natural Earth 110m)
- **Heraldry badges** — each kingdom's heraldry is base64-embedded and placed at the centroid of its largest territory polygon
- **Coordinate recovery** — parks missing `latitude`/`longitude` have their coordinates recovered by parsing the stored `google_geocode` blob (273 parks plotted vs 263 with direct coords only)
- **Freeholds excluded** — kingdom_id 8 (The Freeholds of Amtgard) is not used as a Voronoi node

## Running

```bash
pip install -r map/requirements.txt
python3 map/build_map.py
# Output: map/amtgard_kingdoms.html
```

Requires the `ork3-php8-db` Docker container to be running. The generated HTML file is self-contained and opens in any browser.


baltinerdist:

Synced up to `master` @ `6ab028a3` (merge commit `4683399d`). This branch was **1254 commits behind** — the stalest of the batch.

**Conflicts: 1 file, resolved.**

- **`.gitignore`** — two hunks, both additive on each side, resolved as a union. Kept upstream's `build/` and its tooling-ignore block (`.ork3-db.local`, `tools/ork-db/*`, `tools/fuzzy-validator/*`, `__pycache__/`, `*.pyc`, `.coverage*`, `htmlcov/`, `.pytest_cache/`, `test-results/`) alongside this branch's `map/amtgard_kingdoms.html`. Nothing dropped from either side.

**Method:** `git merge upstream/master`, not a rebase. This branch's history sits behind prior merge commits; a linear rebase would replay them against the new base and manufacture conflicts that aren't real.

**Verification:** no conflict markers in the tree; `php -l` clean across all 375 changed PHP/TPL files pulled in by the merge; `class.Authorization.php` byte-identical to upstream. `git diff --stat upstream/master HEAD` confirms this branch still contributes exactly its own 3 files (`.gitignore`, `map/build_map.py`, `map/requirements.txt`) — feature intact, nothing of upstream's lost.

⚠️ **Worth knowing before this is reviewed:** `map/build_map.py` reads the ORK database directly and was written against a February-era schema. 1254 commits of schema churn just landed underneath it (event RSVP, tournament, qual tables, and more). The *merge* is clean, but the script's queries may be stale. That's functional drift, not a merge problem — it wasn't run as part of this sync.

baltinerdist:
### Schema audit after the 1,254-commit merge

Following up on the caveat in the sync comment above. **Verdict: the script still runs. The merge broke none of its queries.**

The parse path was verified end-to-end against the live DB using stdlib only (nothing installed): **272 usable parks across 26 kingdoms, 0 Unicode decode errors, 4 parks skipped for missing coordinates.**

## Every schema reference is still valid

The script's entire DB surface is a single query.

| Reference | Status | Notes |
|---|---|---|
| `ork_park`, `ork_kingdom` | Valid | survive utf8mb4 + markdown overrides |
| `p.park_id`, `p.name`, `p.kingdom_id` | Valid | |
| `p.latitude`, `p.longitude` | Valid | still `double NOT NULL` |
| `p.google_geocode` | Valid | still `longtext` |
| `p.active = 'Active'` | Valid | still `enum('Active','Retired')` |
| `k.kingdom_id`, `k.name` | Valid | |
| `k.kingdom_id != 8` (Freeholds) | Valid | id 8 confirmed = "The Freeholds of Amtgard" |
| **utf8mb4 conversion** | Changed — but it *helped* | `overrides/2026-03-16-utf8mb4-conversion.sql` converts both tables. The script's `.decode()` defaults to utf-8, so the conversion **fixed a latent latin1 hazard**. 0 decode errors across 276 rows. |
| markdown description columns | No impact | only touches `description`/`directions`, never selected |

The event RSVP, tournament, qualification, custom-title and danger-audit tables the merge introduced simply don't intersect this query. Nothing was renamed, removed, or type-changed in a way that matters.

External deps also check out live: the naciscdn Natural Earth URL returns HTTP 200 and the shapefile still exposes a `NAME` field with exact values `United States of America` and `Canada`.

## Pre-existing issues worth knowing (none merge-induced)

1. **`requirements.txt` has zero pins.** Nothing is stale because nothing is specified — but there's no reproducibility, and it silently opts into whatever geopandas ships next.
2. **`GeoSeries.unary_union` (line 261) is deprecated** in geopandas 1.0 in favour of `union_all()`. Confirmed *not yet removed* as of 1.1.x, so it works today — but `warnings.filterwarnings('ignore')` at line 19 masks the FutureWarning, so this will break with no warning when it's finally dropped.
3. **`--raw` output is fragile by accident.** 144 of 276 rows store real newlines in `google_geocode` (the other 132 store escaped `\n`), so 276 rows emit **18,036 stdout lines**. It survives only because the first fragment of each row still has exactly 7 tab fields and the `int()` guard rejects stray fragments.
4. **`geodatasets` is imported and never used** (line 27) — dead import, still carried in requirements.
5. **Two heraldry badges silently missing:** kingdom 17 has only `0017.png` (the script looks for `.jpg` only), and kingdom 40 has no heraldry file at all. Both render no badge without complaint. Note the upstream masters/WebP image pipeline did **not** restructure `assets/heraldry/` — files are still flat `{id:04d}.jpg`, so that path is intact.

## Not verified

The geo stack was **not executed** — nothing was installed, and the host Python has none of these packages. So the Voronoi → shapely → folium rendering path is unverified by execution. This audit covers data extraction and API compatibility, not a completed render.

## If you keep this open

No changes are needed for correctness. Two cheap hardening items: pin `requirements.txt`, and swap `.unary_union` → `.union_all()` before geopandas removes it.

PR #455, opened 2026-04-09
Experiment: Heatmapping

baltinerdist:
## Summary
- Adds a population heatmap to the **Kingdom profile map tab** — park pins are weighted by active players (1+ sign-ins in the past 12 months), with a toggle between **By Participation** (visitors) and **By Residents** (home park members)
- Adds a global **Atlas Heatmap** page (`Route=Atlas/heatmap`) — same heatmap concept applied across all parks worldwide, gated to kingdom officers and ORK admins
- Atlas index shows a **"Population Heatmap" button** for eligible users, linking to the dedicated page
- Atlas heatmap includes a **Hide Pins** toggle for a clean heatmap-only view

## Auth
Access is gated by querying `ork_authorization` directly for any `kingdom_id > 0` row — correctly handles monarchs regardless of session `kingdom_id` state.

## Test plan
- [ ] Log in as a kingdom officer/monarch — confirm "Population Heatmap" button appears on the Atlas page
- [ ] Navigate to `Route=Atlas/heatmap` — confirm map loads with heatmap overlay
- [ ] Toggle By Participation / By Residents — heatmap density should change live
- [ ] Toggle Hide Pins — park pins should disappear, leaving only heat signatures
- [ ] Click Show Pins — pins restore
- [ ] Click a park pin — sidebar shows Visitors and Residents counts
- [ ] Log out or use a non-officer account — confirm button is hidden and `/heatmap` redirects to Atlas
- [ ] Kingdom profile Map tab — confirm heatmap overlay renders, toggle works


baltinerdist:

Synced up to `master` @ `6ab028a3` (merge commit `12704983`). This branch was **943 commits behind**.

**Conflicts: 2 files, resolved.**

**1. `orkui/controller/controller.Kingdom.php`** — this one needed real work, not a merge choice. Upstream has since moved all inline SQL out of the controller into `Model_KingdomProfile` → `system/lib/ork3/class.KingdomProfile.php`, and added `ParentKingdomId`/`ParentKingdomName` at exactly the spot where this branch had ~35 lines of raw heatmap SQL. That block relied on `$kid` and `global $DB`, **neither of which exists at that point in upstream's version** — so a naive "keep both sides" would have produced a fatal undefined-variable at runtime, while a clean-looking merge.

Resolved by taking upstream's structure and re-homing the feature into it, SQL preserved verbatim:
- `system/lib/ork3/class.KingdomProfile.php` — new `GetKingdomHeatmapWeights(int $kingdomId): array` holding both queries, following the neighbouring `GetKingdomParkDays` house style (`DB_PREFIX`, `$this->db->Clear()` before each query, `->Next()` iteration)
- `orkui/model/model.KingdomProfile.php` — thin pass-through `heatmap_weights()`
- `orkui/controller/controller.Kingdom.php` — now a single line beside the existing `park_days($kid)` call, where `$kid` is actually in scope

Net effect: same feature, same SQL, now sitting in the layer upstream expects it in.

**2. `orkui/template/revised-frontend/Kingdomnew_index.tpl`** — two hunks, both kept-both:
- KnConfig block: this branch's `heatmapWeights:` and upstream's `principalityIds:` landed on the same line → both retained.
- `<style>` block: upstream inserted a new tooltip / royal-crown / copy-link block plus qual-test card CSS in the same region where this branch had prepended heatmap CSS → all of upstream's CSS kept verbatim, the branch's 5 heatmap rules re-inserted at the top of the Move Player `<style>`.

Auto-merged cleanly with both sides intact: `revised.js` (upstream's map legend + principality pins coexist with this branch's `HeatmapLayer`/`knHeatmapMode`), `Atlas_index.tpl`, `controller.Atlas.php`, `Atlas_heatmap.tpl`.

**Method:** `git merge upstream/master`, not a rebase — this branch sits behind prior merge commits, and a linear rebase would replay them against the new base and manufacture conflicts that aren't real.

**Verification:** no conflict markers; `php -l` clean on every changed `.php`/`.tpl`; `node --check` clean on `revised.js`; `class.Authorization.php` byte-identical to upstream. `git diff --stat upstream/master HEAD` is exactly the 8 heatmap files / 713 insertions — feature intact, nothing of upstream's dropped. Maps loader still requests `libraries=marker,visualization`, so `HeatmapLayer` still resolves.

---

Two things noticed while resolving. **Both pre-existing in this branch, neither caused by the merge** — flagging rather than fixing, since this sync was scoped to conflicts only:

1. **Residents query may not match its comment.** The comment says "signed in anywhere", but the SQL joins `m.park_id = a.park_id`, which counts residents who signed in *at their own home park*. I preserved the SQL exactly as written and made the new docblock describe the actual behavior. If the comment was the intent, that's a real bug worth a separate fix.
2. **Heatmap toggle CSS is light-mode only** — hardcoded `#f7fafc` / `#fff` / `#edf2f7`, which will read wrong under `html[data-theme="dark"]`.

baltinerdist:
### Follow-up: dark mode added, residents query corrected

Both items from the sync comment above, in two separate commits: `f4c5fde9` and `c938f03c`.

## 1. Dark mode for the heatmap (`f4c5fde9`)

**Pattern followed, not invented.** Both `orkui.css` and `revised.css` define an identical `--ork-*` token set twice — `:root` (light) and `html[data-theme="dark"]` (dark) — and page CSS uses **explicit override selectors**, not media queries and not a new token layer. `Kingdomnew_index.tpl` already had 86 such rules under a `DARK MODE OVERRIDES` banner; the new rules match that exactly, including its use of `!important` where needed.

66 lines added across three files. Post-push counts: `Kingdomnew_index.tpl` 93 dark rules (was 86), `Atlas_heatmap.tpl` 28, `Atlas_index.tpl` 4.

Notably, **this branch's entire 490-line `Atlas_heatmap.tpl` had zero dark support** — header, stats bar, controls, mode pills, map/sidebar shells, sidebar park body, legend, and the mobile hint (nested in its own `@media` block) all needed it.

**Two things that are not naive inversions:**
- The light toggle is a *recessed* track (`#edf2f7`) holding a *raised* white pill. Inverting each color independently would have made the active pill darker than its track and destroyed the affordance. Instead the track drops to `--ork-bg` and the pill rises to `--ork-bg-tertiary`, preserving the relationship.
- Light-mode red `#c53030` computes to **2.9:1** on `#2d3748` — below AA. Replaced with `#feb2b2` (6.0:1) and `#fc8181` (4.9:1), matching the existing `.kn-btn-danger` dark convention.

Scope respected: in `Atlas_index.tpl` only this branch's own `.at-heatmap-btn` was touched; upstream's surrounding light-only `.at-*` CSS is untouched. Leaflet popups/tooltips were deliberately left alone — the repo themes Leaflet per-page (`Weather_index.tpl` scopes to `.wx-map`) and Atlas doesn't, so tooltips stay on Leaflet's default white where the branch's `#2b6cb0`/`#718096` remain legible.

## 2. Residents query now counts sign-ins anywhere (`c938f03c`)

```sql
- INNER JOIN ork_mundane m ON m.mundane_id = a.mundane_id AND m.park_id = a.park_id
+ INNER JOIN ork_mundane m ON m.mundane_id = a.mundane_id
```

Everything else verified preserved on the pushed branch: `SELECT m.park_id` and `GROUP BY m.park_id` (**still keyed by home park** — only the attendance side widened), `COUNT(DISTINCT a.mundane_id)`, the 365-day window, `a.mundane_id > 0`, and kingdom scoping via `p.kingdom_id`. House style intact: `$this->db->Clear()`, explicit `->Next()`, `(int)` casts, `DB_PREFIX`.

**No cartesian blow-up:** `ork_mundane.mundane_id` and `ork_park.park_id` are both PRIMARY KEYs, so both joins stay many-to-one. A travelling player now contributes more attendance rows to their group, but `COUNT(DISTINCT a.mundane_id)` still collapses them to one.

⚠️ **This changes what the heatmap displays.** Resident counts go **up** for every park whose residents play elsewhere, and the residents layer will read hotter. No park can go down — the old numbers understated resident populations. Note the page is Memcache-cached (1800s), so the shift won't appear instantly after deploy.

**Related finding:** `Map::GetAtlasHeatmapWeights()` in `class.Map.php` never had the home-park constraint, so the Atlas-wide heatmap was always counting sign-ins anywhere while the Kingdom page counted only home-park sign-ins. The two surfaces were silently disagreeing; that discrepancy is now resolved. They are still not byte-identical — the Atlas version additionally filters `m.suspended = 0 AND m.active = 1`, which the Kingdom version does not — so if you want the two fully reconciled, that filter difference is the remaining gap.

## Verification

`php -l` clean on all four changed files. Indentation checked before editing the PHP (`awk '/^\t/{c++}'` → 0 tabs, already PSR-12), so no reformatting occurred. Staged diff reviewed before each commit.

**Not runtime-tested** — the SQL argument is schema-based rather than empirical, and the dark-mode work was not viewed in a browser; the contrast ratios above are computed, not observed.

baltinerdist:
### Follow-up: Kingdom and Atlas heatmaps now agree (`db314465`)

Closes the divergence noted in the previous comment. `class.KingdomProfile.php` now applies the same player filters Atlas already used; `class.Map.php` is untouched.

```diff
  FROM ork_attendance a
+ INNER JOIN ork_mundane m ON m.mundane_id = a.mundane_id
+     AND m.suspended = 0 AND m.active = 1
  INNER JOIN ork_park p ON p.park_id = a.park_id AND p.kingdom_id = {$kid}
```
(participation gained the join outright; residents gained the filter on its existing join)

Verified post-push: 6 insertions, one file. `GROUP BY a.park_id` / `m.park_id`, `INTERVAL 365 DAY`, `a.mundane_id > 0`, `COUNT(DISTINCT a.mundane_id)` and kingdom scoping all survive. `php -l` clean.

## ✅ Verified against the live database, not just by reading

The post-fix SQL was extracted and run against the real dataset alongside Atlas's queries, comparing per-park counts:

- **Participation: 0 mismatches, all parks, all kingdoms**
- **Residents: 0 mismatches, all parks, all kingdoms**
- For contrast, the **pre-fix** participation query disagreed with Atlas on **207 parks** — the gap was real and is now closed.

## Schema evidence — and why it needed checking

`ork_mundane.active` / `.suspended` are both `tinyint(1) NOT NULL`, and `GROUP BY active, suspended` confirms only `0` and `1` ever appear. So `= 1` / `= 0` is correct here.

That check mattered: **the same column name has three different domains in this schema.** `ork_park.active` is `enum('Active','Retired')` and `ork_kingdom.active` is likewise a string — only `ork_mundane.active` is an int flag. The existing `p.active = 'Active'` in `GetKingdomParkDays` alongside `m.active = 1` in the roster query confirms the split.

On semantics: `Attendance::reactivateInactiveMundane()` flips `active` 0→1 whenever a player signs in, so `active = 0` on a player *with* recent attendance means they were deactivated after that sign-in. The filter is narrow and intentional — not a blanket cull of the ~63% of mundanes sitting at `active = 0`, most of whom the date window already excluded.

**No cartesian blow-up:** `SHOW KEYS FROM ork_mundane` confirms `mundane_id` is PRIMARY (`Non_unique = 0`), so the join is strictly many-to-one. Empirically confirmed by the exact-match result above rather than inflation.

## Effect on displayed numbers

**Kingdom page only — down. Atlas unchanged.** Measured on the production-shaped local dataset:
- Participation: 32,289 → 31,094 (**−3.7%**); 207 of 328 parks change, average −4.9, worst single park −264
- Residents: 18,461 → 17,420 (**−5.6%**)

Since shading normalizes within the view, most parks look similar; the visible shift is in parks carrying an unusual number of suspended or deactivated accounts, which now shade cooler relative to neighbours.

## Correction to my earlier comment

I previously wrote that this page is Memcache-cached for 1800s and the change wouldn't appear immediately. **That was wrong — no cache bust is needed.** `GetKingdomHeatmapWeights` has no `ghettocache` wrapper (unlike its neighbours `player_count` at 600s and `GetKingdomPlayersRoster` at 1200s), and `controller.Kingdom.php:325` calls it straight through. The 1800s `heatmap_v1` cache belongs to **Atlas**, whose weights were not changed, so its contents remain valid. The Kingdom change is visible on next page load.

## One remaining inconsistency, deliberately left

**Kingdom uses `INTERVAL 365 DAY`; Atlas uses `INTERVAL 12 MONTH`.** These agree on every date except when the lookback spans a Feb 29, where they differ by one day. Out of the approved scope (player filters), so untouched — but it's the last divergence between the two surfaces if you want them byte-identical.

Two other asymmetries were examined and intentionally not changed, since fixing them would mean altering Atlas: Atlas joins no park table, so it can emit buckets for `park_id = 0` or defunct parks (harmless — nothing matches them on the map); and residents counts `COUNT(DISTINCT a.mundane_id)` vs Atlas's `m.mundane_id`, which is equivalent because the join equates the columns.

**Not runtime-tested through the app** — no browser load of either page. The SQL itself is live-DB verified as above; what remains unexercised is only the unchanged PHP plumbing around it.

PR #474, opened 2026-05-19
Feature/Amtpride Banners

baltinerdist:
## Summary
- Adds an **Amtpride Nameplate** subsection to the player profile Design modal with 15 pride-flag preset gradients (6-color, Progress, Gilbert Baker 1977, Philly POC-inclusive, lesbian, bi, pan, poly, ace, aro, trans, nonbinary, genderfluid, genderqueer, intersex).
- New `ork_mundane_design.hero_gradient varchar(32)` column stores the selected preset key; gradient definitions live in `system/lib/ork3/pride_gradients.php` so new flags are a one-line addition.
- When a flag is active it renders as a horizontal multi-stop linear gradient on the hero and overrides the existing custom primary/secondary colors. Suspended state still wins. Pride colors render as-is in light **and** dark mode — they're identity-bearing and shouldn't be tinted.

## Files
- `db-migrations/2026-05-19-mundane-design-hero-gradient.sql` — adds `hero_gradient` column
- `system/lib/ork3/pride_gradients.php` — preset definitions (key → name + color stops)
- `system/lib/ork3/class.Player.php` — load/persist/validate + audit log entry
- `orkui/controller/controller.PlayerAjax.php` — accepts `HeroGradient` POST param
- `orkui/template/revised-frontend/Playernew_index.tpl` — subsection UI, swatch picker, hero render, dark-mode styles

## Test plan
- [ ] Run the migration: `docker exec -i ork3-php8-db mariadb -u root -proot ork < db-migrations/2026-05-19-mundane-design-hero-gradient.sql`
- [ ] Open the Design My Profile modal on a player; expand **Amtpride Nameplate**; pick each flag and confirm the preview updates and saves
- [ ] Confirm picking a flag dims/disables the custom color presets, and clicking the "Clear pride flag" button restores them
- [ ] Reload the player profile and confirm the hero renders the selected gradient
- [ ] Switch to dark mode and confirm the flag colors render unchanged (no tinting), and the subsection chrome (border, hover, selected ring) is legible
- [ ] Confirm a suspended player still shows the red suspended hero regardless of selected flag
- [ ] Submit a bogus `HeroGradient` value via the AJAX endpoint and verify it is rejected/coerced to null

## Miscellaneous fixes
- **Universal search — typographic punctuation folding.** Names stored with typographic characters (e.g. "Wolf's Run" with a curly apostrophe `’` U+2019) were missed by the top-bar universal search, because the `utf8mb4_unicode_ci` collation does not treat them as equal to their ASCII forms. The search now normalizes both the typed query and the compared columns via a `$punctFolds` map (curly single/double quotes, em/en dashes, no-break space, small tilde) across parks, kingdoms, units, and players. Accented letters already match under the column collation and are left untouched. Code-only; no DB changes.
  - `orkui/controller/controller.SearchAjax.php`

- **Award entry — Given By is now cross-kingdom.** The legacy award-entry form (`Award/park/{id}` → `Award_addawards.tpl`) scoped the **Given By** autocomplete to the page's kingdom by passing `kingdom_id` to `SearchService::Player()`, which hard-filters on it. Awards are routinely presented by players from other kingdoms (wars, coronations, etc.), so the giver simply could not be found. Given By now passes `kingdom_id: 0` so the search spans all kingdoms; the existing `(KAbbr:PAbbr)` suffix disambiguates same-named personas. **Given To** stays kingdom-scoped — you record awards for your own members. Code-only; no DB changes.
  - `orkui/template/default/Award_addawards.tpl`

- **Add Award modal — Custom Title surfacing + "Alias of" on all modals.** Two related Custom Title improvements to the Add Award modal across the player, kingdom, and park profiles. (1) "Custom Title" now appears at the very top of the **Achievement Titles** dropdown (in addition to Awards), so it's discoverable where titles live; this also makes the Achievements tab appear even when a kingdom has no other titles. (2) The "Alias of" dropdown — previously player-only — is now on the kingdom and park modals: their change handlers distinguish Custom Title from Custom Award via data attributes, relabel the custom-name field, show the alias dropdown for Custom Title, and submit `AliasAwardId`. The alias-options query was extracted from `controller.Player.php` into a shared `Model_Award::fetch_custom_title_alias_options()` used by all three controllers.
  - `orkui/model/model.Award.php`, `controller.Player.php`, `controller.Kingdom.php`, `controller.Park.php`, `Kingdomnew_index.tpl`, `Parknew_index.tpl`, `revised-frontend/script/revised.js`

- **Player search — relevance ranking + hardened `KD:PK` prefix.** A company "Add Member" search couldn't find a cross-kingdom player ("Silent", GP:SKW) without a prefix. `SearchService::Player()` ordered results purely `active DESC, persona`, so a common token's exact match was alphabetically truncated past the row limit. Added an exact/prefix persona relevance tier to the `ORDER BY` (all search types/callers benefit). Also hardened `magic_search`: anchored the prefix regex (matching the universal search) and made the park separator tolerant of a trailing colon, so `GP:SKW: Silent` no longer mid-string mis-matches as `SKW: Silent`. Code-only; no DB changes.
  - `system/lib/ork3/class.SearchService.php`

- **Household "Add Member" search — filter within park instead of an alphabetical roster.** The Unit add-member/add-manager autocomplete was the last player search still on the legacy `SearchService` + `magic_search` path (tiny `LIMIT 8` per tier, fragile 3-call merge). In large parks the typed name failed to narrow results, so the dropdown showed the park roster A–Z truncated at the limit — users couldn't reach later names (e.g. "Wolf"). Ported it onto the canonical `KingdomAjax/playersearch` endpoint, mirroring the Event RSVP modal (`evPnBuildGroups`): LIKE-based filtering, own-kingdom-first ordering, `LIMIT 15` per tier; tiers *In Park* → *In Kingdom* → *All Players* (deduped, header only when the tier has unseen rows); abbreviation prefix (`nb:ff wolf`) handled server-side. Also hardened: persona rendered via `textContent` (was raw `innerHTML`) and a sequence guard so a slow earlier response can't overwrite a newer keystroke.
  - `orkui/template/default/Unit_index.tpl`

- **Award Recommendation modal — mobile viewport access.** On phones the recommendation / add-award modal had no `max-height`, so the vertically-centered box pushed its header (title + close ✕) off the top and its footer (Submit/Cancel) off the bottom with no way to scroll to either. The base `.pn-modal-box` is now viewport-bounded (`max-height: calc(100dvh - 32px)`, flex column) with a pinned header/footer and an internally-scrolling body — the same pattern the reconcile modal already used; the Kingdom/Park modals already shipped this cap. The rank number buttons also shrink ~30% (36px → 25px) at ≤600px so the rank grid stops dominating the modal. CSS-only; no DB changes.
  - `orkui/template/revised-frontend/style/revised.css`

- **Rank pill selector — unified earned / granted / selected scheme.** Redesigned the rank pills shared by every add-award and recommend surface (player, kingdom, park; add, recommend, edit, reconcile) into one identical scheme driven by a shared painter: earned ranks render green with a faint ✓ watermark, ranks being granted render blue with a faint → watermark, and the chosen rank pops up (blue, ↑ watermark, colored glow ring). Selecting a rank above the player's current rank lights every step forward (right arrows) up to the chosen rank (up arrow). Removes the prior per-profile / Kingdom-green divergence, enlarges and bolds the number for legibility over the watermark, and ships full dark-mode parity. The Rank helper text now explains that green ranks are already awarded and that a higher rank may be suggested. CSS + JS; no DB changes.
  - `orkui/template/revised-frontend/style/revised.css`, `script/revised.js`, `Playernew_index.tpl`, `Kingdomnew_index.tpl`, `Parknew_index.tpl`

- **Mobile nav — fixed top bar clipped on Android Chrome.** The global top bar (`#newmenu`) is `position: fixed; width: 100%`, and the root had no horizontal-overflow guard. Any descendant wider than the viewport makes the page scroll sideways, and on Android Chrome a fixed `width:100%` bar then spans the full *scrollable* width — pushing its right-side controls (Login / View Profile / theme toggle) off the right edge. iOS clamps fixed elements to the visual viewport, so it never reproduced there. Added `overflow-x: hidden; max-width: 100%` to `html, body` so the fixed nav stays pinned to the device width on every platform; both `position: sticky` elements in the app stick within their own scroll containers, so they're unaffected. (Symptom fix — the specific overflowing descendant is still to be pinpointed via a mobile emulator.) CSS-only; no DB changes.
  - `orkui/template/default/style/orkui.css`

- **Park Day edit — description silently truncated.** The park-day edit modal (`Parknew_index.tpl`) lets users enter up to 200 characters, but `ork_parkday.description` was `varchar(50)`. With `sql_mode` empty the DB silently truncated anything past 50 chars on save — no error, the save reported success and the page reloaded, so longer-description edits appeared to do nothing. This was the sibling column missed by `2026-03-29-markdown-description-columns.sql` (which widened the description columns on `ork_park` / `ork_unit` / `ork_event_calendardetail`). Widens the column to `varchar(200)` to match the modal's input limit. Already-truncated existing rows can't be recovered; fixes edits going forward.
  - `db-migrations/2026-05-21-widen-parkday-description.sql`

- **Design modal — live heraldry-overlay preview.** The **Heraldry Overlay Strength** buttons (Low / Medium / High / Vignette) in the Design modal previously only set a hidden field — they had no visible effect on the modal's hero preview, and the preview never rendered the player's heraldry, so all four strengths looked identical until the profile was saved and reloaded. The preview now carries the same blurred heraldry layers the production hero uses, driven by a preview-scoped `--pn-preview-overlay-opacity` var (so the real hero's `--pn-overlay-opacity` is untouched). A small `applyOverlayPreview()` helper sets the opacity / toggles vignette compositing, wired into the button clicks, the Reset button, and modal init so the preview opens reflecting the saved strength. Frontend only; `HeroOverlay` already persists, so no DB or controller changes.
  - `orkui/template/revised-frontend/Playernew_index.tpl`
  - Design spec: `docs/superpowers/specs/2026-05-21-heraldry-overlay-preview-design.md`

- **Award recommendation +1 — recommendations cache now keyed by viewer.** `PlayerAwardRecommendations` embeds per-viewer fields in its response (`ViewerCanSecond`, `ViewerCanEditReason`, and each second's `IsMine`), but the GhettoCache key was keyed only on `KingdomId/ParkId/PlayerId` — it omitted `RequestedBy`. So the first officer to load a player's recommendations after a cache bust had **their** eligibility flags cached and served to every viewer for the 300s TTL. Concretely: a recommender of a ladder rung holds an own-rec for that `(player, award, rank)`, so their computed `ViewerCanSecond` is false for every rec on that rung; when that response was cached and served to another officer, the +1 to second a co-recommendation disappeared — the officer could second one of two recs for the same rung but not the other. Adds `RequestedBy` to the cache key so each viewer gets their own correctly-computed flags (writes already flush memcache, so contaminated entries clear on the next second/withdraw or the TTL). Code-only; no DB changes.
  - `system/lib/ork3/class.Report.php`

- **QA hardening pass.** A two-cycle multi-agent QA review of the branch surfaced and fixed a set of issues across the features above: moved the Custom Title alias-options raw SQL out of the `model` layer into `class.Award` (lib layer) to honor the project's DB-in-lib separation; fixed a `pk`→`kn` CSS-prefix mismatch in the Kingdom recommendation rank pills; restored visible dark-mode hover feedback on rank pills (the prior box-shadow was invisible dark-on-dark); replaced hardcoded inline grays in the new "Alias of" dropdown rows with dark-mode-aware CSS variables; hardened the new search-relevance `ORDER BY` interpolation (the codebase's `mysql_real_escape_string` is a no-op polyfill); collapsed dead-branch hint ternaries; and added a defensive null guard plus missing `name` attributes. Code-only; no DB changes.
  - `system/lib/ork3/class.Award.php`, `class.Player.php`, `class.SearchService.php`, `model.Award.php`, `Playernew_index.tpl`, `Kingdomnew_index.tpl`, `Parknew_index.tpl`, `revised-frontend/script/revised.js`, `style/revised.css`

### DIFF
--- a/db-migrations/2026-05-19-mundane-design-hero-gradient.sql
+++ b/db-migrations/2026-05-19-mundane-design-hero-gradient.sql
@@ -1,0 +1,11 @@
+-- Amtpride Nameplate: stores a preset key (e.g. 'pride6', 'progress', 'trans')
+-- that the player template renders into a horizontal multi-stop gradient on
+-- the hero. When set, takes precedence over color_primary/color_secondary.
+-- Key validation lives in class.Player.php::UpdatePlayer against the keys
+-- defined in system/lib/ork3/pride_gradients.php.
+
+ALTER TABLE `ork_mundane_design`
+  ADD COLUMN IF NOT EXISTS `hero_gradient` varchar(32)
+  CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci
+  DEFAULT NULL
+  AFTER `color_secondary`;
--- a/db-migrations/2026-05-21-widen-parkday-description.sql
+++ b/db-migrations/2026-05-21-widen-parkday-description.sql
@@ -1,0 +1,9 @@
+-- Widen ork_parkday.description from varchar(50) to varchar(200).
+-- The park-day edit modal (Parknew_index.tpl) allows maxlength=200, but the
+-- column was varchar(50). With sql_mode empty, longer descriptions were
+-- silently truncated to 50 chars on save, so edits appeared to "do nothing".
+-- This was the sibling missed by 2026-03-29-markdown-description-columns.sql,
+-- which widened the description columns on ork_park / ork_unit / ork_event_calendardetail.
+ALTER TABLE `ork_parkday`
+  MODIFY COLUMN `description` varchar(200)
+  CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '';
--- a/docs/superpowers/specs/2026-05-21-heraldry-overlay-preview-design.md
+++ b/docs/superpowers/specs/2026-05-21-heraldry-overlay-preview-design.md
@@ -1,0 +1,96 @@
+# Live Heraldry-Overlay Preview in the Design modal
+
+**Date:** 2026-05-21
+**Surface:** Playernew "Design My Profile" modal → Colors tab
+**Type:** Enhancement (frontend only)
+
+## Problem
+
+The Design modal's hero preview (`#pn-color-hero-preview`) live-updates for color
+and gradient choices, but the **Heraldry Overlay Strength** buttons (Low / Medium /
+High / Vignette) have no visible effect there. They only set the `pn-hero-overlay`
+hidden field. The preview box also never renders the player's heraldry, so all four
+strengths look identical. Users cannot tell how their heraldry will bleed through the
+hero until they save and reload the profile.
+
+## Goal
+
+Make the overlay-strength buttons (and the modal's initial state) update the preview
+in real time, compositing the player's heraldry over the chosen color/gradient exactly
+as the saved profile hero will render it.
+
+## Constraints
+
+- Frontend only, confined to `orkui/template/revised-frontend/Playernew_index.tpl`.
+- No DB or controller changes — `HeroOverlay` already persists and round-trips.
+- Must match production hero compositing so the preview is truthful.
+- Must not alter the real hero's rendering or its `--pn-overlay-opacity` variable.
+
+## How production composites the hero (reference)
+
+`revised.css:166-202` + `Playernew_index.tpl:951-956`:
+
+- `.pn-hero` paints the base color/gradient (`--pn-hero-bg`).
+- `.pn-hero-bg` is an absolutely-positioned heraldry layer, `background-size:cover`,
+  `filter:blur(6px)`, `opacity: var(--pn-overlay-opacity)`.
+- Opacity map: `low 0.06`, `med 0.12`, `high 0.22`.
+- Vignette mode uses two layers — `.pn-hero-bg-vignette-base` (heavy blur, masked OFF
+  at center) + `.pn-hero-bg-vignette-sharp` (light blur, 0.30 opacity, masked ON at
+  center) — for a sharp center fading to a soft, low-opacity edge.
+- Heraldry layers render regardless of pride-gradient mode (they sit on top of the
+  pride background too).
+- Heraldry URL is already available server-side as `$heraldryUrl`; players with no
+  custom heraldry get the `000000.jpg` placeholder (same as production).
+
+## Design
+
+### 1. Preview markup
+Add heraldry layers as the first children of `#pn-color-hero-preview`, mirroring the
+real hero structure, using `$heraldryUrl`:
+- `.pn-hero-bg pn-preview-bg pn-preview-bg-vignette-base`
+- `.pn-preview-bg-vignette-sharp`
+
+Both vignette layers stay in the DOM; a `pn-preview-vignette` class on the preview box
+controls whether normal or vignette compositing is active. The preview content
+(name + subline) keeps `position:relative; z-index:1` so it stays above the layers
+(the `.pn-hero-preview` rule already sets `position:relative; overflow:hidden`).
+
+### 2. Preview-local opacity variable
+The preview reuses the production layer CSS (blur, masks) but is driven by its **own**
+opacity variable so the real hero is untouched. The preview-scoped `.pn-preview-bg`
+rule reads `var(--pn-preview-overlay-opacity, 0.12)`, set inline on the preview box by
+JS. Vignette base/sharp layers use preview-scoped selectors so their opacity/blur/mask
+match production (base uses the preview opacity var; sharp is fixed at 0.30 like prod).
+
+### 3. Wire-up (existing handlers)
+- **Overlay buttons** (`tpl:3941`): on click, in addition to the existing active-class
+  + hidden-field update, set the preview's `--pn-preview-overlay-opacity` from the
+  same map, toggle `pn-preview-vignette` for the vignette button, then call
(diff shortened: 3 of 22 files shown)


baltinerdist:
### Amtpride banners — hero text legibility over light flag gradients

Several pride flags (trans, ace, the white/yellow bands of the 6-color, etc.) render light gradient stops, which washed out the white/translucent hero text. Added a `pn-hero-pride` hook (set on `.pn-hero` server-side when a gradient is active, and toggled on the design-modal preview in `updateColorPreview()`) and scoped these overrides to it:

- **Persona name** — drop shadow darkened `rgba(0,0,0,0.4)` → `0.6` (~20%).
- **Subline (mundane name) + breadcrumb** — near-opaque white (`0.95`) + matching `0.6` shadow. `text-shadow` inherits, so pronouns, pronunciation guide, and separators are covered.
- **Breadcrumb links** → solid white; **bullet `·` / chevron `›` separators** → full opacity.

Applies in light **and** dark mode (the gradients are light in both); non-pride profiles are untouched, and suspended state still clears the gradient upstream so the red signal wins.

- `orkui/template/revised-frontend/Playernew_index.tpl`

PR #475, opened 2026-05-26
Enhancement: Crown RBAC and Officers Module

baltinerdist:
## Overview

Makes kingdom officer positions **first-class, kingdom-extensible, alias-able, RBAC-bound data**, built on a new role-based access control (RBAC) foundation. Replaces the hardcoded five-role officer model with a position registry that supports Crown vs Supporting classification, custom positions, per-kingdom title aliases, nesting ("reports to"), retire/reinstate, and per-position permission binding.

A full design spec and phased implementation plan are included under `docs/superpowers/`.

## RBAC foundation
- New tables: `ork_permission`, `ork_role`, `ork_role_permission`, `ork_user_role`, `ork_rbac_audit` (system + kingdom-custom roles, scoped grants, audit).
- `PermissionRegistry` (atomic permissions) + `RBACService` (`HasPermission`, role CRUD, officer→role sync, self-appointment guard).
- `Authorization` bridge (`HasPermissionOrAuthority`) for gradual migration from legacy auth.

## Officers module
- **Position registry** (`ork_officer_position` + `ork_officer_position_alias`): `ork_officer.role` migrated ENUM→VARCHAR + `position_id` FK with dual-write/backfill; Core-Five seeded and pinned.
- **Manage Officers UI**: a card under Kingdom Admin → modal with Crown/Supporting cards, set/vacate occupant, create/edit position, reclassify, retire/reinstate, retired list. Fully dark-mode.
- **Title aliasing everywhere**: code keys on a stable `canonical_key`; all UI renders an alias-aware `DisplayTitle` (per-kingdom aliases for the shared Core-Five via the alias table).
- **Occupancy rules**: one office holds one person, for both classifications, serialised by an advisory lock keyed on the office. A person may hold **any number** of offices, at any number of levels — an earlier per-person Crown limit was removed after it was found to refuse 242 people who currently hold office (see the transition section below).
- **Nesting**: optional `parent_position_id` ("Reports To") with indented display and cycle/scope validation.
- **Hide-when-vacant**: non-Crown positions can be hidden from read-only displays when empty.
- **RBAC binding per position**: choose an existing role, build a custom permission set, or **None** (recorded/displayed, no extra permissions). Pinned Core-Five keep their system role.
- Two distinct permission gates: `kingdom.officer.set` (appoint) vs `kingdom.officer.position.manage` (restructure the registry).

## Hardening (review/polish pass)
- Cross-kingdom IDOR guards on all position write paths + role-rebind ownership validation.
- Self-appointment null-sentinel guard; orphan-role cleanup; PHP-8 `mysql_real_escape_string` fatal fixed.
- Performance: in-PHP cycle walk, aggregated escalation check, batched role-binding/permission writes.
- Maintainability: deduped Kingdom/Park `GetOfficers`, single-cased response shape, `@deprecated` markers on the legacy role-map path.
- **Player search fixes**: corrected malformed search URLs (`&q=` not `?q=` — `UIR` already ends in `?Route=`, so a second `?` emptied `$_GET['q']` and returned no results), scoped the officer searches to the kingdom, defined the previously-undefined `tnFixedAcPosition` dropdown positioner, and matched result-row formatting to the shared award-modal style.

## Officer writes are now a real API surface

Every officer write was taking the acting user's identity as a **parameter** and relying on a
controller for authorization. That is why none of them could be exposed, and why `RBACService`'s
`GrantRole` escalation guard — which compares the granter's own permissions — was comparing the
permissions of whoever the caller *claimed* to be.

All of them now match the shape the rest of this application already uses (compare
`Player::AddAward`): a `$request` array carrying a `Token`, the actor resolved from
`IsAuthorized($request['Token'])`, a scope-correct permission gate **inside the domain method**, and
a `dangeraudit` row.

- **`OfficerPosition`** — 12 gated methods: `TransitionOfficer`, `SetOccupant`, `VacateOfficer`,
  `VacateOffice`, `CreatePosition`, `EditPosition`, `ReorderPositions`, `RetirePosition`,
  `ReinstatePosition`, `AddHistoryTerm`, `EditHistoryTerm`, `DeleteHistoryTerm`. Registered in
  `orkservice/Json/index.php`.
- **`RBACService`** — the five role mutators are gated but the class is deliberately **not**
  registered; its 21-method read surface has had no authorization pass.
- **Scope-correct permission keys.** `PermissionRegistry` has defined `park.officer.set`,
  `park.officer.vacate` and `park.officer_history.manage` all along and nothing used them; the admin
  mapped every action, including vacate, to `kingdom.officer.set`. `park.officer.position.manage` was
  referenced by no code at all.
- **`EditHistoryTerm`/`DeleteHistoryTerm` authorize against the row's own kingdom and park**, not the
  caller's claim — otherwise anyone administering any kingdom could edit another kingdom's rolls by
  naming their own.

### A transition is now a recorded event
`TransitionOfficer` closes the outgoing officer's term at a **caller-supplied, possibly backdated**
date, opens the incoming one, and persists the note the admin console has been collecting and
discarding since it shipped. The whole read-close-move-open span is serialised under an advisory lock
keyed on the office; two admins transitioning the same office previously produced two rows with
`end_date IS NULL`, which reads as an office held by two people.

### The rolls have data
`ork_officer_history` held **one row against 2,507 seated officers**, so the rolls rendered empty
everywhere and no sitting officer had an open term. The backfill opens one per seated officer.

**Every backfilled `start_date` is NULL.** Using `ork_officer.modified` was considered and rejected:
of the 66 rows with a "usable" value, 42 share a single date across 42 *distinct parks* and 17 share
another across 17 — bulk row-creation artifacts, not people taking office. Commit `52f729f7` on this
branch had already reached that conclusion. Real start dates will be filled in through the UI.

### Two exposure fixes worth calling out
`JsonServer::validate_method` applies no method whitelist — it rejects a requested name only if it
starts lowercase, contains `_`, or is `__construct`. Registering a class therefore publishes **every**
public PascalCase method on it. Casing is not a filter, because PHP method names are case-insensitive
and the caller picks the casing; only an underscore or `private` actually hides a method.

- `GetOfficersForDisplay` returns `GivenName`/`Surname` unconditionally, while the public-page query
  `Kingdom::buildOfficerRows` blanks them for anonymous viewers. It is now `get_officers_for_display`.
- `Kingdom::buildOfficerRows` was itself dispatcher-reachable on the already-registered `Kingdom`
  class, taking the caller-supplied `$is_authorized` flag. Now `build_officer_rows`.

`tests/Unit/ApiExposureTest.php` fails the build if an ungated public method appears on a registered
class. It is scoped to `OfficerPosition`; other registered classes remain unswept.

### Known follow-ups
Recorded in `docs/superpowers/specs/2026-08-29-officer-transition-design.md`. The load-bearing one:
the officer admin controller's outer gate is still hardcoded to kingdom scope, so a park-only officer
is refused before reaching the (correct) domain gate. **No UI surface was removed in this pass** —
the edit pads and legacy pages still work, and consolidating them is the follow-on plan.

## Migrations (apply in order)
`rbac-tables.sql` → `rbac-seed.sql` → `officer_history.sql` → `rbac-migrate-officers.sql` → `officer-position.sql` → `officer-position-nesting.sql` → `officer-position-alias-sort-order.sql` → `officer-history-backfill.sql`

The final backfill is pure data (no DDL) and should run in a maintenance window — it takes no officer locks.

## Testing
This branch now carries an automated suite: **762 main + 159 schema-tooling tests, zero failures**, run via `bin/run-unit-tests.sh` (which also gates on `ork-db drift-check --strict`). Officer coverage lives in `tests/Integration/Officer*Test.php`, `RbacAuthorizationTest`, and `tests/Unit/ApiExposureTest`.

Additionally verified by hand against a populated dev database: the Manage Officers UI flows (create/alias/set/reclassify/retire/nest/None), cross-kingdom IDOR rejection, player-search endpoints returning scoped results, and officer rendering on kingdom & park profiles. The history backfill was rehearsed against a clone of a production backup, twice, to confirm idempotency.


baltinerdist:

Synced up to `master` @ `6ab028a3` (merge commit `eec24d2c`). This branch was **518 commits behind** with **29 conflicting files** — the heaviest in this batch. Now 0 behind / 34 ahead.

## Root cause of the conflict storm

Upstream did a repo-wide PSR-12 reformat (tabs→4 spaces) **and** the "megiddo" architecture refactor: controllers no longer call `Ork3::$Lib->authorization->HasAuthority()` — they call `$this->Authorization->has_authority()` (model → new `AuthorizationGate` facade) — and auth logic was pulled *out of templates* into controller-supplied flags. Every one of this branch's `HasAuthority` → `HasPermissionOrAuthority` substitutions landed on a line upstream had moved or rewritten.

**Method:** for each conflicted file, upstream's refactored version was taken as the base and this branch's semantic changes re-applied on top, then verified by **call-site accounting — the branch had 33 RBAC call sites; the merged tree has 33 equivalents, mapped 1:1.**

`class.Authorization.php` did not conflict (upstream never touched it since the merge base); the branch's version survives byte-for-byte. `HasPermission` and `HasPermissionOrAuthority` both present, **no `true ||` bypass** (independently re-verified).

## Integration work that was required

Upstream's facade had no counterpart to this branch's RBAC bridge, so thin passthroughs were added rather than bypassing the new layer:
- `class.AuthorizationGate.php` — `checkPermission()` / `checkPermissionOrAuthority()`, delegating to `Authorization`
- `model.Authorization.php` — `has_permission()` / `has_permission_or_authority()`

## Resolutions

**14 controllers** — all "upstream's refactored body + this branch's permission keys", using the new `$this->Authorization->has_permission_or_authority(...)` form. Keys re-applied across `Search`, `Principality`, `Tournament`, `Unit`, `Attendance`, `Reports`, `Kingdom`, `Park`, `Event`, `EventAjax`, `ParkAjax`, `KingdomAjax`, `Player`, `Admin`. Every check survived.

Two notable ones:
- **`controller.Event.php`** — where upstream and this branch *both* changed a check (event `edit`/`reconcile`, `CanManageEvent`/`CanManageAttendance`), upstream's legacy role + staff-delegation clause was kept and the branch's key added. Reverting to the branch's version would have **deleted upstream's event-staff feature**.
- **`controller.EventAjax.php`** — 3 gates (`add_attendance`, `delete_rsvp`, `heraldry`) had moved upstream into `EventPlanning->can_*()` staff-aware gates. Upstream's gate was kept and `has_permission(...)` OR'd in, so neither side's check was dropped and the legacy `HasAuthority` isn't run twice.

**6 lib classes** — `class.Attendance.php` (4 keys), `class.Heraldry.php` (9), `class.Event.php` (8), `class.Player.php` (29, mapped line-by-line against upstream's relocated code), `class.Kingdom.php` (12), `class.Park.php` (12). Also ported: the branch's officer-position JOIN into upstream's extracted `GetOfficerRoles()`, `buildOfficerRows()` shared by Kingdom+Park, `SetOfficer`/`VacateOfficer` position wiring + `changed_by`, and the 4 `*OfficerHistory` methods.

Two **non-conflicted** files needed the feature re-homed because upstream had extracted the exact code this branch modified: `class.KingdomProfile.php::AuthorizeMovePlayer()` → `player.move` (where upstream moved KingdomAjax's move-player gate).

**9 templates** — upstream deleted inline auth calls in favour of controller-supplied flags. Each branch permission key was re-homed into the controller/helper that now computes the flag; **every check survived, none dropped, only relocated**. `Kingdomnew_index.tpl`/`Parknew_index.tpl` were purely additive on both sides (branch officer-history CSS/JS + upstream dark-mode/copy-from-past-event CSS and username-availability JS) — both kept and confirmed present post-merge.

## ⚠️ Things that need a human decision

1. 🚨 **The qualification gate may lock out kingdom officers.** This branch's hardening dropped the `AUTH_KINGDOM` and standalone `AUTH_PARK` branches, leaving:
   ```php
   HasPermissionOrAuthority($requester_id, 'player.qualification.edit', 'park', $this->mundane->park_id, AUTH_EDIT)
     || HasAuthority($requester_id, AUTH_ADMIN, 0, AUTH_EDIT)
   ```
   The legacy fallback is now **park-scoped only**, so a kingdom officer without park-scoped authority loses access unless the RBAC seed data grants it. Preserved as-is because it is this branch's deliberate hardening commit, not a merge artifact — but it needs confirming against real role data before merge.
2. **`AuthorizationGate` is arguably where `HasPermission`/`HasPermissionOrAuthority` belong.** Upstream built that facade precisely *because* `class.Authorization.php` is local-only — and `.githooks/pre-commit` line 28 does `git reset -q HEAD "$AUTH_FILE"`, unstaging it from **every** commit. These two methods already exist in history so they ship fine today, but **any future edit to them will be silently unstaged** and appear to vanish. Moving them into the facade would remove that trap.
3. **`Admin::kingdom()` `CanAddPark`** is `has_authority($uid, AUTH_ADMIN, (int)$id, AUTH_CREATE)` — `AUTH_ADMIN` with a *kingdom* id. `HasAuthority` ignores the id for `AUTH_ADMIN`, so this reads as a per-kingdom check that isn't one. Looks like a pre-existing branch bug; preserved verbatim rather than "fixed" during a merge. (Related: the same confusion was just resolved on #496 — park creation is global-admin only.)
4. **Other branch tightenings preserved as-is** (deliberate hardening, not merge artifacts): `Player::UpdatePlayer` top gate `AUTH_EDIT`→`AUTH_CREATE`; `active`/`park_member_since` `AUTH_EDIT`→`AUTH_CREATE`; `SetPlayerSuspension` re-scoped kingdom→park.
5. **`Event`/`EventAjax` role tuning** — where upstream and branch disagreed on `AUTH_CREATE` vs `AUTH_EDIT`, upstream's (newer) was kept, since the branch had only changed the function name there, not the role. Worth a glance if that wasn't the intent.
6. **`Kingdom::SetKingdomAwards` is commented out upstream.** The branch's `kingdom.award.edit` key was applied inside the comment block for fidelity — dead code either way.

## Redundancy to be aware of

Upstream's `EventPlanning::CanManageEventDetail`/`CanAddAttendance`/`CanRemoveRsvp` and `KingdomProfile::AuthorizeMovePlayer` are extracted gates duplicating what this branch patched inline. Upstream's event-staff delegation (`event_staff.can_manage/can_attendance/can_schedule/can_feast`) is a **parallel, non-RBAC permission system** covering some of the same ground as this branch's `event.*` keys — worth deciding whether they converge. Upstream's `set_admin_kingdom_auth_flags()`/`admin_can_edit_kingdom_reports()` overlap the branch's `Admin::kingdom()` auth-flag block; both now run.

## Verification

No conflict markers; no `true ||` bypass; `php -l` clean on all **314** changed `.php`/`.tpl`. Merge recomputed from its parents with `rerere` disabled and diffed against what was pushed — 32 files, exactly the 29 conflicts plus the 3 disclosed integration edits, nothing unexplained. (Two other branches in this batch were contaminated by a stray line from an unrelated `rerere` resolution; this one is clean.)

**Not smoke-tested:** requires this branch's migrations (`officer-position.sql`, `officer_history.sql`, `rbac-*.sql`); the local DB lacks `ork_officer_position*`, `ork_officer_history`, `ork_role`, `ork_user_role`. **Those migrations must be applied before this ships.**

**PSR-12:** `model.Kingdom.php` and `model.Park.php` merged cleanly with the branch's tab-indented methods appended to upstream's space-indented files — mixed style, no functional impact, the pre-commit hook will normalize on the next real edit.

baltinerdist:
### Follow-up: RBAC/officer-admin DB access moved into the domain layer (`d0db08c8`)

Separate commit, kept apart from the sync merge so each stays reviewable.

**Distinct query sites: 3, not 13.** The 13 grep-matching lines collapse to 3 real queries across 4 locations — the large assignments query was byte-for-byte **duplicated** in both controllers.

| # | Query site (before) | Domain method (`class.RBACService.php`) | Controller call site |
|---|---|---|---|
| 1 | `ork_user_role` ⨝ `ork_role` ⨝ `ork_mundane` + LEFT JOIN granter, kingdom-scoped, non-expired — `Admin.php:1803` | `GetKingdomRoleAssignments($kingdom_id, $with_park_names = false)` | `Admin::roles()` |
| 1' | *same query, duplicated* — `KingdomAjax.php:808` | same method, `$with_park_names = true` | `KingdomAjax::rbac()` → `getassignments` |
| 2 | `SELECT COUNT(*) FROM ork_user_role WHERE role_id = …` — `Admin.php:1856` | `GetRoleUserCount($role_id)` | `Admin::roles()` |
| 3 | per-row `SELECT name FROM ork_park WHERE park_id = …` inside the loop — `KingdomAjax.php:844` | folded into site 1 behind `$with_park_names` | — |

Site 3 was folded rather than exposed separately so the exact query *sequence* is preserved — the N+1 park lookup still fires per park-scoped row, on the same `$DB` handle, with the same `Clear()`. The flag exists because only the AJAX endpoint emitted a `ParkName` key; `Admin`'s `RoleAssignments` array keeps its original key set.

**No new domain class.** `system/lib/ork3/class.RBACService.php` already exists (added by this branch) and unambiguously owns `ork_role`/`ork_user_role`/`ork_role_permission`, with sibling read helpers (`GetUserRoles`, `GetRolePermissions`, `GetAvailableRoles`) in its "ROLE QUERY HELPERS" section — so the two new methods went there rather than splitting one domain across two classes. A thin `orkui/model/model.RBACService.php` (21 lines) was added, copied structurally from `model.OfficerPosition.php`, relying on `Model::__construct` auto-wiring and `Model::__call` forwarding. Name resolution verified: `Model_RBACService` → `RBACService` → `DIR_ORK3 . class.RBACService.php`, which `startup.php` requires.

Per the note in my sync comment above, **nothing was added to `class.Authorization.php`** — the pre-commit hook unstages it from every commit, so anything placed there would silently fail to ship.

**Verification:** both controllers now have **zero** SQL and **zero** `$DB` references (no stray `global $DB;` either); `php -l` clean on all 4 changed files; no `true ||` bypass. SQL equivalence was proven by extracting the original and new query strings into a standalone harness and comparing them whitespace-normalized — all three report **IDENTICAL** (the literals differ only in indentation, since the SQL was re-indented to match RBACService's tab style).

**Deliberately not moved:** the controllers still call `Ork3::$Lib->rbacservice->GetAvailableRoles()`, `GrantRole()`, `RevokeRole()`, `CreateRole()` etc. directly, bypassing the model. Those are **domain calls, not raw DB access** — converting them would turn a pure refactor into a call-convention migration. Worth a follow-up consistency pass.

**Not smoke-tested:** the local MariaDB has no `ork_role`/`ork_user_role` tables (`SELECT COUNT(*) FROM ork_role` → `ERROR 1146`) — this branch's RBAC migrations have never been applied to the shared local DB, and loading them would have been a destructive change to a resource other branches share. Verification is lint + SQL-string equivalence + return-shape inspection. **Someone with the RBAC schema loaded should smoke-test `Admin/roles/Kingdom/{id}` and `KingdomAjax/rbac/{id}/getassignments` before merge.**

baltinerdist:
### Follow-up: qualification gate restored, RBAC checks moved to the facade

Both open items from the sync comment above, in two separate commits: `b8930821` and `54d30cbf`.

## 1. Qualification gate — kingdom fallback restored (`b8930821`)

```php
// now
if (Ork3::$Lib->authorizationgate->checkPermissionOrAuthority($requester_id, 'player.qualification.edit', 'park', $this->mundane->park_id, AUTH_EDIT)
    || Ork3::$Lib->authorization->HasAuthority($requester_id, AUTH_KINGDOM, $this->mundane->kingdom_id, AUTH_EDIT)
    || Ork3::$Lib->authorization->HasAuthority($requester_id, AUTH_ADMIN, 0, AUTH_EDIT)) {
```

The kingdom leg is **verbatim from upstream, not invented** — `git show upstream/master:system/lib/ork3/class.Player.php` carries the pre-hardening gate with exactly `HasAuthority($requester_id, AUTH_KINGDOM, $this->mundane->kingdom_id, AUTH_EDIT)`. `$this->mundane->kingdom_id` is a real property used in 10 places in the same file. Nothing was removed: the park leg still lives inside `checkPermissionOrAuthority`'s legacy fallback (`'park'` → `AUTH_PARK`). Kingdom officers can no longer be locked out by missing RBAC seed data.

## 2. RBAC checks moved to `AuthorizationGate` (`54d30cbf`)

Implementations now live in `class.AuthorizationGate.php` as `checkPermission()` / `checkPermissionOrAuthority()`. The wrappers added during the sync were **collapsed into** the real logic rather than duplicated, so there is exactly one implementation each and no delegation loop.

**99 call sites across 12 files** updated (Player, Park, Kingdom, KingdomProfile, Unit, Event, Heraldry, Attendance, Treasury, Tournament, Report, controller.OfficerAdminAjax). There were **zero** `authorization->HasPermission(` call sites — every external use went through the OrAuthority bridge. Verified post-push: no reference to the old call path remains anywhere in `*.php` / `*.tpl` / `*.js`.

`Ork3::$Lib->authorizationgate` is auto-registered by `startup.php`'s `scandir` loop, same as every other lib class. `model.Authorization.php` **needed no edit** — its `has_permission` / `has_permission_or_authority` already called `_authorization_gate()->checkPermission*()`, and those now resolve to the gate's own implementation.

### 🔐 Two deliberate choices worth your attention

**The lowercase `check*` names are a security decision, not a style one.** `JsonServer::validate_method()` rejects methods starting with a lowercase letter, and both `Authorization` and `AuthorizationGate` are exposed at `/Json`. Naming these `HasPermission*` on the gate would have **published a new arbitrary-`mundane_id` privilege oracle** over HTTP. Please keep them lowercase.

**Parameters were left untyped** on the moved methods (the earlier `int $mundaneId` / `string $scopeType` hints were dropped). With 99 call sites now routed through them, a hinted `int` would turn any `null`/`''` mundane id into a fatal TypeError that the old code tolerated — so this keeps the move strictly behavior-preserving. The `scope_type → AUTH_*` switch became a lookup array with the same `AUTH_PARK` default (verbatim semantics; done because php-cs-fixer wanted to mangle the inline-`break` switch).

### What happened to `class.Authorization.php` — and the one call needing your sign-off

**It was not touched at all.** It is byte-identical to `origin`, and `grep -n 'true ||'` on it is empty.

Its `HasPermission()` / `HasPermissionOrAuthority()` are now **dead code**, which is safe unconditionally:
1. **No redeclare** — they're members of `Authorization`; the new copies are members of `AuthorizationGate`. Each declared once, in its own file.
2. **No shadowing** — `AuthorizationGate extends Ork3`, *not* `Authorization`. There is no inheritance relationship in which one could override the other.
3. **No reachability** — every caller now names `Ork3::$Lib->authorizationgate` explicitly.
4. The only remaining path to the old copies is the **pre-existing** public `/Json?call=Authorization/HasPermissionOrAuthority` endpoint, unchanged from before this PR.

They were deliberately **not deleted.** Because these commits use `--no-verify`, a deletion *would* have shipped — but it would also write a change into the one file the repo is engineered to keep out of commits, guaranteeing a conflict against every maintainer's local override of that file. The dead copies are the safer end state; retiring them belongs with retiring the hook. **That's the one judgment call here worth your explicit sign-off.**

## Verification

`php -l` clean on all 14 changed files; no `true ||` bypass; exhaustive grep for the old call path returns nothing; php-cs-fixer reports `class.AuthorizationGate.php` conforming. (`controller.OfficerAdminAjax.php` is flagged, but that's pre-existing — the whole file is tab-indented — and was left alone rather than sweeping a 500-line reformat into this diff.)

**Not runtime-tested.** PHPUnit isn't installed in the worktree (`vendor/` gitignored), so `tests/Integration/AuthorizationLibTest.php` wasn't run. Two things worth exercising against a real DB with the RBAC migrations applied: (a) a kingdom officer *without* park authority saving reeve/corpora quals, and (b) any RBAC-gated save, to confirm `Ork3::$Lib->authorizationgate` resolves at runtime.

baltinerdist:
### Follow-up: dead RBAC permission oracle removed (`1c7a5a73`)

Resolving the sign-off item from the previous comment — and on investigation this was a security fix, not just tidiness.

**Why it mattered.** `checkPermission*` were given deliberately lowercase names on `AuthorizationGate` because `JsonServer::validate_method()` (`class.JsonServer.php:308`) rejects any method whose first character is lowercase. But the abandoned copies in `class.Authorization.php` started with uppercase `H` and contained no underscore, so they **passed** that check — and `orkservice/Json/index.php` lists `'Authorization'` in the allowlist. That left `/Json?call=Authorization/HasPermissionOrAuthority` reachable as an arbitrary-`mundane_id` privilege oracle, defeating the precaution entirely. Leaving them was not neutral.

Registration was irrelevant to the exposure: `JsonServer::RunServer()` → `call_endpoint()` → `run_call()` does `class_exists($class)`, `new $class()`, then `call_user_func_array` — no registration entry required. (Confirmed anyway that `AuthorizationService.registration.php` never declared either method, so nothing needed removing there.)

**Confirmed dead before deleting.** A tree-wide grep across `*.php`, `*.tpl`, `*.js`, `*.json`, `*.ini`, `*.sql`, `*.html` found **zero** `Authorization->HasPermission(` / `Authorization::HasPermission` call sites, including `orkservice/` and `orkui/`. The only matches were the definitions themselves, `RBACService`'s own `HasPermission()` (not in the `/Json` allowlist), and `AuthorizationGate:34` calling `rbacservice->HasPermission(...)`. Two markdown planning docs mention the name in prose.

Also confirmed strictly branch-own code: `git log -S` traces both to `59a50eb7` ("Add RBAC bridge methods to Authorization class") on this branch, and `git merge-base --is-ancestor 59a50eb7 origin/master` says not on master.

**Verified post-push:** the commit is 41 contiguous deletions and nothing else — `git show --stat` lists the file (the `--no-verify` path defeats the hook's `git reset`, which was checked rather than assumed). Zero `function HasPermission` declarations remain, zero `true ||` bypass, the four `KeyExists` guards around line 329 are intact, and `php -l` is clean.

**Gate exposure audit — clean.** Of `AuthorizationGate`'s five public methods, only `HasAuthority(array $request)` passes `validate_method`, and it is already hardened: it resolves the actor from `$request['Token']` via `IsAuthorized()`, returns `BadToken()` on failure, and **ignores any client-supplied `MundaneId`**. It answers "am *I* authorized", not an arbitrary-subject query. `check`, `checkPermission`, `checkPermissionOrAuthority` and `clearSharedDb` are all lowercase-blocked; the `Ork3` base contributes only `__construct`/`__set`, both blocked.

---

## ⚠️ Separate, pre-existing, and NOT introduced by this branch

While verifying the above I traced the `/Json` dispatch path fully, and there is an equivalent exposure on the **legacy** method that predates this PR. Flagging it rather than touching it, since it is upstream code with a public API surface.

`Authorization::HasAuthority($mundane_id, $type, $id, $role)` is public, uppercase-initial, underscore-free, and on an allowlisted class. The dispatcher resolves arguments by **reflection** when no explicit definition exists — `get_default_definition()` reflects the method's parameters, and `wrangle_parameters()` pulls each one by name straight from the request into a positional `call_user_func_array`. `RunServer()` → `call_endpoint()` → `run_call()` performs **no authentication check** in that path; only `valid_class` and `valid_method` are consulted.

The practical effect is that `/Json?call=Authorization/HasAuthority&mundane_id=…&type=…&id=…&role=…` appears to be an **unauthenticated** query against any player's authority, with the subject chosen by the caller. That is information disclosure (who holds what authority) rather than privilege escalation, since the method returns a boolean and grants nothing — but it is worth a deliberate look, and the fix pattern already exists in `AuthorizationGate::HasAuthority`, which takes a token and ignores caller-supplied ids.

Not verified by an actual HTTP request — this is a read of the dispatcher, so please confirm before acting on it.

**Not runtime-tested.** The change is a pure deletion with zero resolved call sites, but a smoke test of the officer-admin surfaces before merge would confirm it.

baltinerdist:
Pushed four commits. Summary for reviewers, with the security- and data-affecting items first.

### Fixes that affect live data

These were found by auditing the Kingdom Admin modals and are independent of any UI change:

- **Every player move cleared that player's waiver.** `MovePlayer` read a `Waivered` request key that none of its five callers passed, so it always evaluated to `0`. Waivers are per-kingdom, so it now clears only on a cross-kingdom move and leaves a within-kingdom move untouched.
- **Officer-created accounts with a blank password could be logged into.** `CreatePlayer` salted `USERNAME + ''` and login checks the identical expression. An empty password now creates no credential at all. Self-registration already enforced 8 characters; officer creation enforced nothing.
- **Vacating a supporting office deleted every holder** via a `DELETE` with no `mundane_id` filter, behind a confirm written in the singular. It now ends the term in officer history rather than deleting service records, and supports vacating one holder.
- **`CreateAward` always reported success** — it returned nothing and the controller treated `null` as a pass, so a rejected insert showed "Award created!" and reloaded. The same truthiness idiom was swept from three other actions.
- **`RBACService::IsAdmin()` accepted any `role='admin'` row**, including a park-scoped one, bypassing the all-zero-scope guard in `HasAuthority`. It gated the global permission bypass plus the escalation guards on Grant/Create/Edit/DeleteRole.
- **Retiring a park through Edit Parks skipped `kingdom.park.retire`** and the danger audit, because the Active toggle went through `SetParkDetails`.

### Destructive actions now state their blast radius

Award delete became a **soft delete** on the existing `ork_kingdomaward.disabled` column (no migration). Deleting *Order of the Rose* would have orphaned **3,772** granted awards; the confirm now says so, disabled awards cannot be granted, and they can be re-enabled. Reset Waivers reports its count and points at the audit log. Merge Players previews both records with a swap control and commits behind a five-second wait-to-arm button.

### Console and modals

The Admin console's four stat cards were rendering `$TrendStats`, which only `Admin::index()` assigns — they showed `0` with no delta. They're now a work queue reading kingdom-scoped counts, each linking to the report that lists the set. Six template variables the page read but nothing ever assigned are now populated, so heraldry, the hero counters and the Parks admin table work. Manage Awards gained search, alphabetical order and collapsed groups (127 rows was roughly eight screens); `Peerage` now reaches the template, which fixes grouping that was filing 17 Paragons under Masterhoods. A new `ConfigRegistry` supplies officer-facing labels and typed controls for all 13 kingdom config keys, replacing raw database keys in the UI.

Shared modal layer: dialog semantics and focus management, a dirty guard on every close path, topmost-only Escape, and `tnFixedAcPosition` hoisted out of a permission-gated partial that three ungated modals depended on.

### Deferred — needs a product decision

Two sets of controls are written and read by nothing, so they're flagged rather than changed:

- **Park Titles**: `minimumattendance`, `minimumcutoff`, `period`, `period_length` — no park promotion or demotion code exists anywhere in the repo.
- **Manage Awards**: `reign_limit` / `month_limit` — no award-granting path reads either.

Both need a call on whether to implement the behaviour or remove the inputs.

### Verification

`php -l` clean across all changed files, layering rules R1–R7 clean, no SQL or domain instantiation in the presentation layer, no Smarty leaks, no native dialogs, every added colour rule has a dark-mode counterpart, and the unit suite is at the pre-existing **184 / 427 / 4 errors / 17 failures** baseline on every commit.

baltinerdist:
### Update: officer transition backend (22 commits)

Pushed the officer-transition backend work and refreshed the description. Three statements in the
old body were no longer true and have been corrected: the one-Crown-office-per-person rule (removed),
"Supporting = unlimited holders" (an office now holds one person), and "no automated test harness
exists in this codebase" (there are now 762 + 159 tests).

**What changed, in one line:** every officer write became a token-gated `$request` domain method with
the actor resolved from the token rather than passed as a parameter, `OfficerPosition` is registered
on the JSON API behind a regression guard, and `ork_officer_history` — which held one row against
2,507 seated officers — is backfilled.

Three findings from the review pass are worth surfacing on their own, because each was a live defect
rather than a cleanup:

**The one-Crown-office-per-person rule refused 242 sitting officers.** All five system positions are
seeded `classification = 'crown'` and 705 parks reuse those same shared rows, so the rule treated a
park's Champion exactly like a kingdom's Monarch. Measured against a production backup: 176 people
hold two offices in the same park, 59 across parks, 30 hold both a park and a kingdom office. Zero
hold two kingdom-level Crown offices. The ORK does not limit holding multiple offices, so the check
was deleted rather than narrowed.

**`GetOfficersForDisplay` would have published real legal names.** It returns `GivenName`/`Surname`
unconditionally, while the parallel public-page query `Kingdom::buildOfficerRows` takes an
`$is_authorized` parameter — docblock: *"Whether the requester may see private name fields"* — and
blanks them for anonymous viewers. Registering `OfficerPosition` with that method exposed would have
made every seated officer's legal name anonymously readable over HTTP. Both it and
`buildOfficerRows` are now underscored, which is what actually hides a method from the dispatcher —
a lowercase initial does not, since `method_exists` is case-insensitive and the caller picks the
casing.

**The backfill would have written fabricated dates.** The original design used `ork_officer.modified`
as the start date where "usable" — 66 of 2,507 rows. Of those, 42 share one date across 42 *distinct*
parks and 17 share another across 17: bulk row-creation artifacts, not people taking office. Commit
`52f729f7` on this branch had already concluded a backfill "has to treat pre-existing rows as
start-date-unknown rather than reading this column." Every backfilled `start_date` is now NULL.

**Verification.** Full suite green (762 main + 159 schema-tooling, zero failures, `drift-check
--strict` passing). The migration was rehearsed against a clone of a production backup: 2,507 seated
officers went from 0 to 2,507 open terms, the single future-dated row was reopened, no officer ended
with two open terms, and a second run produced byte-identical counts. All 12 mutators return
`Status 5` unauthenticated over HTTP.

**No UI surface was removed.** The edit pads, the legacy set-officers pages, and the history controls
on the profile modal all still work. Consolidating them into the admin area is the follow-on plan,
and it is blocked on one thing recorded in the spec: the officer admin controller's outer gate is
still hardcoded to kingdom scope, so a park-only officer is refused before reaching the domain gate
that now handles park scope correctly.

PR #485, opened 2026-06-17
Enhancement: Discord QOL — editable Player Since override + mobile rec-button spacing

baltinerdist:
Two Discord-requested quality-of-life items.

## 1. Editable "Player Since" override (admin)

"Player Since" on the player profile was always computed as `MIN(ork_attendance.date)`, so it could never predate a player's first *recorded* sign-in — players with incompletely-imported early attendance looked more recent than they actually were.

This adds an admin-editable override:
- New nullable column `ork_mundane.player_since_override` (migration in `db-migrations/`). `NULL`/zero-date = use the computed value; a date overrides it. Mirrors the existing `park_member_since` override.
- New lib resolver `Player::get_player_since_date()` — the override-vs-computed coalesce lives entirely in the library layer (no business logic leaks into controllers).
- A "Player Since (override)" date field in the Update Account modal (admin-only), with a "Current first sign-in: …" hint.
- When set, the override drives all three surfaces that read the value: the profile "Player Since" row, the Amtgard Tenure widget, and the "First sign-in at Amtgard" milestone.

**Bugfixes found during end-to-end testing:**
- yapo's `YapoSave` silently skips `null`-assigned fields (`isset()` guard), so clearing the override never persisted. Fixed by writing the zero-date to clear (the resolver treats it as "use computed"). The same latent bug on the adjacent `park_member_since` clear is fixed too.
- Added server-side `checkdate()` validation so a malformed/impossible date can't be stored.

**Migration ordering:** run the migration *before* deploying the code — under PDO `ERRMODE_WARNING` an unknown column silently rolls back the entire mundane `UPDATE`.

Verified end-to-end locally: set → profile shows the override; clear → reverts to computed; invalid input rejected; unrelated fields preserved.

## 2. More vertical space between stacked rec action buttons (mobile)

On the award-recommendations tabs (Player, Park, Kingdom), the Grant / Delete / +1 buttons stack vertically on mobile (≤768px) and sat only 5px apart, causing mis-taps. Bumped the stacked gap to 10px. Single shared `.pk-rec-actions` rule covers all three surfaces; desktop layout and dark mode unaffected.

---

Opened as a **draft**. Design/plan notes are included under `docs/superpowers/` — happy to drop those from the PR if you'd prefer a leaner diff.


baltinerdist:

Synced up to `master` @ `6ab028a3` (merge commit `302fed89`). This branch was **518 commits behind**.

**Conflicts: 3 files / 57 hunks, resolved. The PR diff shrank from ~11,000 lines to 656.**

**Root cause of the conflict storm:** two commits on this branch (`747464a5`, `2c04f1f8`) ran php-cs-fixer over the *entirety* of `class.Player.php` and `controller.Admin.php` — ~10,800 changed lines wrapped around a ~40-line feature. Upstream reformatted the same files independently. So nearly every hunk conflicted on **formatting, not meaning**.

Before resolving, the reformat was proven to carry zero semantic content: both files were token-normalized (`token_get_all`, whitespace and comments stripped) at merge-base and branch-tip and diffed. The entire non-feature delta was `{`/`}` placement, `else if`→`elseif`, `NULL`→`null`, `Array`→`array`, and one dropped `?>`. Given that, the correct resolution was to take upstream's file wholesale and re-apply only the feature.

- **`system/lib/ork3/class.Player.php`** (27 hunks) — reset to upstream, three feature edits re-applied verbatim: `'PlayerSinceOverride'` in the player-details array; the `UpdatePlayer` park-auth block (zero-date clear for `park_member_since` + validated `$pso` persist via `preg_match`/`checkdate`, `'0000-00-00'` on empty/invalid); and `get_player_since_date()` placed at its original neighbour above `get_earliest_park_attendance_date`.
- **`orkui/controller/controller.Player.php`** (1 hunk) — reset to upstream, `PlayerSinceDate` + `PlayerSinceComputed` re-applied at both call sites.
- **`orkui/controller/controller.Admin.php`** (29 hunks) — reset to upstream, single `'PlayerSinceOverride'` passthrough line re-applied.

Auto-merged clean and verified feature-intact: `Playernew_index.tpl` (override `<input type=date>` + hint, still inside the `$canEditAdmin` guard) and `revised.css` (`.pk-rec-actions` mobile `gap: 14px`).

**An upstream refactor this branch absorbed:** the third `controller.Player.php` hunk was a comment above inline milestone-building code. Upstream has since moved that logic into `Player::get_player_milestones()`, which takes `PlayerSinceDate` as an input. Since the controller now feeds it the *resolved* date, the First Sign-In milestone follows the override exactly as designed — no code needed, comment obsolete. No feature was dropped.

**Verification:** no conflict markers; `git ls-files -u` empty; `php -l` clean on all 297 changed `.php`/`.tpl` files; `class.Authorization.php` byte-identical to upstream with no bypass. `git diff --stat upstream/master` = 8 files, +656/−4 — feature only. Independently confirmed post-push that all three `class.Player.php` edits are present in the pushed file (lines 959, 2816, 4596).

---

**Four things for a human before this merges:**

1. ⚠️ **This branch's php-cs-fixer reformat of those two files was deliberately discarded** in favor of upstream's formatting. Token-proven to be zero behavior change, and it's why the diff went from 11k lines to 656 — but it *is* a decision, so confirm you're happy losing that reformat.
2. 🚨 **Migration `db-migrations/2026-06-17-add-player-since-override.sql` is still unapplied** anywhere refreshed from prod since June. `ork_mundane.player_since_override` must exist before this ships.
3. **`get_player_milestones()` internal fallback** (`class.Player.php:1308`) still calls `get_earliest_attendance_date()` rather than `get_player_since_date()` when a caller omits `PlayerSinceDate`. The player-profile path always passes the resolved value, so the override is honored there — the fallback was left alone to avoid a drive-by that could break `tests/Unit/PlayerMilestonesTest.php`. Worth a decision.
4. **New interaction that didn't exist when this branch was written:** upstream's Qualifications modal posts to the same `update_player` endpoint with hidden passthrough fields and does *not* include `PlayerSinceOverride`. Traced `Request::__get()` — a missing key returns `null` and the lib guards on `!is_null($pso)`, so saving that modal won't wipe the override. Safe, but worth knowing.

PR #486, opened 2026-06-23
Enhancement: Project Front Door

baltinerdist:
## Overview

Introduces a new public **"front door"** landing page for the ORK, plus a **v2 content-management system (CMS)** that powers it and a family of content pages and a blog.

## Front Door

- Cinematic, photo-rich, newcomer-first landing that replaces the utilitarian home page — hero **carousel**, live **stat ticker**, education sections (*What is Amtgard / Find Your Path / Your First Day*), upcoming events, photo mosaic, kingdoms teaser, and a closing call-to-action.
- Marketing nav mirroring amtgard.com (now CMS-editable).
- **Adaptive:** logged-in members get a slim personalized bar; anonymous visitors get the full marketing experience. Dark-mode + responsive throughout.
- The previous dashboard becomes a standalone **Kingdoms Directory** at `Directory/index` (linked from the nav).
- Rendered from a content-block model (`Model_FrontDoor` → `frontdoor/render_blocks.tpl` → `frontdoor/blocks/{type}.tpl`).

## CMS (v2)

- **Universal block renderer** — the front door, CMS pages, and blog entries all render as an ordered list of blocks. "Page type" is an editor preset, not a separate renderer.
- **Page types:** composed / article / media / resource / blog / dynamic.
- **20+ block types:** hero carousel, rich_text, card_grid, gallery (lightbox), video_embed, accordion, table, columns (recursive), file_download, cta_band, image, quote, heading, divider, plus dynamic feeds (events, kingdoms, blog).
- **Admin editor** (`Cms/index`): page & post management, drag-orderable block editor with **TinyMCE**, **media library + image picker**, draft / publish / **preview** / autosave.
- **Blog:** posts + tags, paginated index, entry pages, **RSS**.
- **Editable navigation** (the marketing nav reads from the CMS).
- **RBAC:** cumulative roles (contributor → author → editor → publisher → admin), `CmsCan()` with a bridge to the existing `HasAuthority` so kingdom/park officers gain rights when scoped pages arrive. A gated **"Manage Site Pages"** entry appears in the user drop-down.
- **Security:** strict HTML sanitization (DOM allowlist) on all authored content; URL/scheme validation on links and nav.
- **Scope-ready** data model (`scope_type`/`scope_id` — global now; kingdom/park drop-in later with no migration).

## Data / migrations

- New `ork_cms_*` tables (`db-migrations/2026-06-23-cms-foundation.sql`) + idempotent seed scripts (home page, marketing nav).
- New web-served `assets/cms-media/` dir (runtime uploads gitignored; `.gitkeep` tracked).
- The front door is served from the CMS store, falling back to hardcoded defaults when unseeded.

## Quality

Built and verified phase-by-phase, then put through a multi-lens **polish pass** (security, data-integrity, stability, performance, conventions) with implementer→critic review:
- XSS hardening (sanitizer URL bypasses, nav-link validation, recursive nested-field sanitization).
- Data integrity (dup-slug `lastInsertId` guard via read-back; `ReplaceBlocks`/`Reorder` transactions).
- Stability (session-token check no longer mass-logs-out on a transient DB error).
- Perf (capability resolution collapsed; blog tag N+1 → single batched query; RSS caching).
- DRY `CmsBase` refactor across the CMS libs.

Specs: `docs/superpowers/specs/2026-06-22-amtgard-front-door-design.md`, `docs/superpowers/specs/2026-06-23-amtgard-cms-design.md`.

---

### CMS Theme Engine
Token-based theming for the public CMS — admins restyle colors, typography, and shape/density from curated **design tokens** (plus an advanced raw-token panel), no CSS. Values are stored per scope and injected as a `:root { --token: value }` block at render; front-door stylesheets reference `var(--token)`. **Dark mode is auto-derived** from the single palette (WCAG-contrast-checked). Live editor with preview + light/dark toggle. No free-form CSS (values only flow into named custom properties).

### CMS Multi-Site — per-kingdom public websites
Cashes in the scope-ready data model: **each kingdom gets its own standalone public website** at `/k/{slug}`, authored with the same block builder.

- **Public routing + standalone chrome** (`Controller_Site`): own header/logo, scoped nav, own theme, own footer with a subtle "Part of the Amtgard ORK ↗" tie-back. Per-org theme injected from the site's own scope.
- **Scope-aware admin:** the existing CMS editor threads a validated `?scope=` selector (re-checked server-side via `HasAuthority` — never trusted). Officers self-serve edit; monarch/regent (`AUTH_ADMIN`) publishes. **IDOR guard on every mutation.** Unpublished sites 404 to the public but **preview** for authorized officers, who also get the edit/new-post FAB.
- **Org-scoped dynamic blocks:** live Officers, Parks (sortable by name/city/state, optional heraldry + rank badges), Events, and an interactive **Parks Map** (the Amtgard Atlas map + click-to-open detail sidebar) scoped to the kingdom.
- **Starter template** seeded on first "Manage Public Site" (Home / About / Our Parks / Officers / Documents + nav, published-ready); Directory **"Visit site"** discovery link.
- **Global Sites overview** (`Cms/sites`, ORK super-admins): Amtgard International pinned on top, then every started kingdom/park site — status badge, page/post counts, drill-in to the scoped CMS admin, inline **publish/unpublish**, and **new-site provisioning**.
- **Block picker** refresh: wide **3-column** layout with per-block descriptions and **collapsible groups**.

Data model is park-ready (`/p/{slug}` deferred). Spec: `docs/superpowers/specs/2026-07-01-cms-multi-site-kingdom-sites-design.md`.

### Latest CMS-module polish
Multi-lens review (72 sharded reviewers) of the CMS module, verified/fixed: **`IsSafeUrl` percent-encode XSS bypass** closed (`%6aavascript:` / `java%09script:`, tested), park-name escaping in the map block, media-search placeholder fix, **`DeletePage`/`DeletePost` atomicity** (read-back + rollback) + opt-in scope-ownership guards, **`RevokeRole` fail-closed** auth, and block-render fault isolation (one bad block can't blank the page).

---

## Update — Distributed review, then a mobile & accessibility pass

Three further rounds since the last update.

### Distributed review — problem findings
Multi-lens review of the whole CMS/front-door platform, each shard verified by an adversarial critic pass.

- **Persistence integrity** — block-save verification compares every column plus an order-insensitive JSON check; zero-row / trashed `UpdatePage`/`UpdatePost` now fail loudly; the dup-slug guard distinguishes a real collision from a dropped write; `RestoreRevision` hard-fails on a corrupt snapshot.
- **Routing** — slug/parent redirects with prefix fallback, post-rename 301s, leaf-only publish gating, cycle and cross-scope parent rejection, transliterated slugs.
- **Caching / perf** — GhettoCache on the home page + blocks and site-by-slug, gated `_promoteScheduled`, an in-memory admin path map (kills an N+1), clamped and cached teaser.
- **Multi-site scope** — fail-closed scope normalization, global grants keyed at 0, `RevokeRole` capability probe, scoped person lookup, nav that resolves only live targets.
- **Security** — sanitizer size/depth budget, recursion depth cap on `columns`, media where-used scan that includes embedded ``, `copyDoc` MIME allowlist, CSRF on preview-publish.
- **Migrations** — relative UIR seeds, full demo chain, guarded FK adds, idempotent content seeds, hard-required sanitizer.

### Distributed review — 24 enhancements
- **Front door** — `content-visibility` on block sections, block surfaces routed through `--fd-*` theme tokens (zero visual change), carousel ARIA region + live announcements, per-block `<style>` dedup, gallery lightbox scroll-lock + inert background, preview-only block-failure placeholder.
- **CMS lib** — shared `CmsPage::HTML_FIELDS`/`URL_FIELDS`, content-mutation audit trail, `PurgeTrashed` + orphan-block sweep + unused-tag prune, block-only saves bump owner `updated_at`/`updated_by`, DB-enforced one-active-theme-per-scope.
- **Controllers** — memoized block catalog, batched bulk media delete, centralized page-type enum, per-view analytics deferred off the render critical path, `clearrendercache` + `runmaintenance` actions.
- **Editors** — shared `cms-admin.js` (one source for toast / modal / CSRF-post), view counts on Pages & Posts plus a top-content dashboard panel, `window.CMS_CAPS`.
- **Multi-site** — single grant-driven edit-FAB helper across Page/Blog/Site, `sitemap.xml`.
- **Tests** — CmsSanitizer XSS regression, CmsThemeTokens math, block-field allowlist contract.

### Mobile & accessibility remediation
A mobile-review pass over the front-door blocks, the CMS admin shell, and the shared theme chrome.

- **Layout** — CTA rows wrap via flex/gap instead of `margin-left` on the nth button; the steps grid steps 4→2→1 column; the path grid collapses at 520px; carousel dots and the play/pause toggle lift above the 2×2 stat ticker; mosaic tiles take a uniform height with `object-fit:cover`; the member bar actually hides on phones; long org names ellipsize instead of shoving the hamburger off-screen; rich-text and raw-HTML tables scroll in their own container rather than forcing horizontal page scroll; CMS nav-manager rows wrap so actions keep a 40px touch target; the recovery-email input moves to 16px to stop iOS zoom-on-focus.
- **Contrast (WCAG AA)** — `#b8860b` (2.9:1 on white) darkened across kickers, nav hover/active, officer roles, blog cards and the RSS link; muted greys lifted in `file_download` and the parks map. Dark mode added for the recovery-email modal and the parks-map empty states; `--ork-warn` defined globally so the media library's "In use" label is no longer an undefined token.
- **Accessibility** — icon-only controls get `aria-label` with `aria-hidden` on the glyph; the search toggle becomes a real `<button>` and the maintenance banner an `aria-expanded` button; focus trap + focus restore on CMS admin modals and background `inert` on the staff-roster modal; media tiles are keyboard-operable with a focus ring; kingdom-profile tabs get `role=tablist`/`tab`/`tabpanel`; icon-rail labels are visually hidden rather than `display:none` so the accessible name survives; Directory section titles promoted to `<h2>`; scrollable tables get `tabindex=0` + `role=region`.
- **Also** — the carousel supports touch swipe (passive, horizontal-only); CMS/blog pages suppress the stacked app nav; the `Page_view` breadcrumb uses `UIR` instead of a hardcoded path; `photo_mosaic` surfaces an empty-state hint in the CMS preview.

---

## Update — OGRE branding, adversarial review, and park sites

Five commits since the last update.

### The CMS is now "OGRE" (Online Gallery and Resource Engine)
Product name surfaced in the admin chrome and author-facing copy — rail wordmark (with a visually-hidden acronym expansion so a screen reader doesn't read an opaque token), dashboard lede, page titles, deny page, draft-preview banner, nav dropdown, and the kingdom "Manage Public Site" tooltip. `Cms*` classes and `ork_cms_*` tables stay as the internal identifiers.

### Adversarial review — three hostile passes
Core engine, presentation, then a third pass attacking the remediation itself and the test suite.

- **Seeding** — `CmsSite` inferred "starter template never seeded" from live content, so an org that deleted its seeded nav links or trashed a starter page got them silently re-created **as published pages**. Replaced with an explicit `template_seeded_at` marker plus a separated repair path. The schema probe guarding that column used `try/catch`, which is unreachable under PDO `ERRMODE_WARNING` — now `SHOW COLUMNS`.
- **Slugs are addresses** — `_normalizeSlug` used `iconv('ASCII//TRANSLIT')`, whose output is libc- and locale-dependent: the same org name derived `creconom` in the container and `cr-econom` on a macOS host, so one kingdom could get two different live URLs depending on which machine created the row. Replaced with an explicit transliteration map + NFD normalization (a decomposed paste converges on the same slug as the precomposed form), verified byte-identical on both hosts. `CmsPost::_slugify` carried a second copy of the same call — tag slugs are both a URL segment and the dedup key, so one tag name entered on two hosts produced two `ork_cms_tag` rows; now routed through the shared helper.
- **Scope / tenancy** — `mediabulkdelete` accepted ids without filtering them to the caller's scope; added `FilterOwnedIds` (int-cast into the `IN` list, never the no-op escape shim) with a cap that fails closed. Media reference counts ignored `deleted_at`, so soft-deleted owners pinned media forever — split into live-owner counting for the reversible guard and all-owner counting for the irreversible purge.
- **Leaks & clocks** — breadcrumbs leaked unpublished ancestor slugs, and the due-check compared PHP's `America/Chicago` clock against a UTC database, keeping live ancestors redacted for ~5h after publish. `is_live` is now computed in SQL.
- **Presentation** — the block editor built its summary with `innerHTML` on a *detached* div, and a detached `` still fires in Chrome; replaced with a DOM-free `stripHtml()` that decodes entities in a single pass so `&amp;lt;` cannot cascade into a real `<`. That table had also been folded to lowercase two lines under the comment forbidding it, flattening exactly the Norse and Old-English names (`&AElig;`, `&Eacute;`) it existed for.
- **Editor / picker** — the home-page picker read a 500-row `updated_at DESC` list and reported "no home page chosen" when the real one aged off the cap (a page nobody edits falls off first); it now injects the current home page on a miss, scope-checked against the save rule.

### Park sites are live
`/p/{slug}` is served alongside `/k/{slug}`. New `park_meeting`, `park_officers` and `park_events` blocks, with kingdom and park officer rendering sharing one partial. `Controller_Site::_prefixFor()` is now the only way to derive an org URL — the site-settings modal had hard-coded `/k/` in three places, showing park officers the wrong public address for their own chapter, and its permission copy named "monarch or regent" in park scope where the gate is `AUTH_PARK`. `CmsSite::OrgUnitNoun()` returns Kingdom / Principality / Park so copy stops hard-coding "kingdom"; principalities need no schema change, being `ork_kingdom` rows with a `parent_kingdom_id`.

### Three new single-source-of-truth classes
Each replaces a pair of declarations that could drift apart silently.

- **`CmsBlockRegistry`** — the one declaration of the block catalog and page-type presets. The editor chooser, starter-block seeds, page-type allowlist and the `CmsAjax` save-time allowlists all read it, so the write side can no longer drift from the editor's vocabulary.
- **`CmsMeta`** — canonical + Open Graph assembly. Five controllers had hand-rolled the same `$PageMeta` literal and their own http/https origin detection.
- **`CmsRenderCache`** — the render-cache key registry. Memcached has no prefix scan, so "flush this org's site" can only be an enumeration; the block partials that *write* keys and the endpoint that *busts* them previously declared namespace, key format and clamp bounds independently, and drift there failed in the worst direction — the endpoint reports "cleared 433 entries" while the officer who just fixed their roster keeps seeing the stale one.

### Front door
The "Record Keeper" nav link is now an **"ORK Login"** outline button (destination unchanged — `Route=Directory`). The label lived in three places that all had to agree, including the `marketing_nav` DB row that the live home page actually renders, so a PHP-only edit would have changed nothing — hence `db-migrations/2026-08-09-cms-nav-ork-login-label.php` (idempotent, leaves deliberately-renamed labels alone, reads each write back since `Execute()` is void under `ERRMODE_WARNING`). Also trimmed ~48px of dead clearance reserved for `#newmenu`, which the front door hides.

### Tests
New `tests/cms-tenancy` covers the multi-tenant boundary in `CmsMedia` — scope filtering, storage metering, upload quota — code three review passes had *read* but nothing had ever *executed*. `site_test.php` was red on this branch (it pinned the pre-transliteration slug); corrected and extended with Æ/Þ/Ø/ß and NFD-vs-NFC cases. **All five suites pass.**

### Also
Fixed a pre-existing master bug where `ENVIRONMENT=DEV` served a **blank 200** at the document root — the `HTTP_UI` redirect sat inside the non-DEV branch only.

Migrations: `2026-08-09-cms-site-seed-marker.sql` (additive, re-run safe), `2026-08-09-cms-nav-ork-login-label.php`.

---

## Update — OGRE in-browser capabilities test, and the six fixes it produced

The CMS declares 32 block types and **8 of them had never had a single row written
to the database** (`spacer`, `table`, `video_embed`, `blog_feed`, `raw_html`,
`park_meeting`, `park_officers`, `park_events`), so no amount of static review could
reach them. Every addable block type was instantiated through the Add-block chooser,
saved, and looked at in Chrome across three kitchen-sink pages — global, kingdom, and a
park site provisioned during the run. Plan and full results:
`docs/superpowers/plans/2026-08-10-ogre-capabilities-test.md`.

**30/30 addable block types pass.** Also verified: park-site provisioning, scope gating
(`kingdom_*` correctly absent from the park chooser), the page-type chooser, autosave,
media upload (master + thumb + WebP rendition, all served), dark/light contrast (0
failures), 390px horizontal overflow (0 offenders), and 0 console errors on public pages.

The XSS battery — entered through the real editor save path, not injected into the DB —
was fully neutralized: `<script>`, ``, `<svg onload>`, `javascript:`,
`%6aavascript:`, `java&#9;script:` and a foreign `<iframe>` all stripped, benign markup
intact, the three anchors keeping their text and losing their `href` entirely.

### Six findings, all fixed in this branch

- **Park sites were seeded with the KINGDOM starter template** (high). A new park site
  got `kingdom_events`, `kingdom_officers`, `kingdom_parks` and `kingdom_parks_map` —
  each of which correctly renders *nothing* outside a kingdom scope — an "Our Parks"
  page for an org with no parks, and copy calling the park a kingdom. The blocks and the
  chooser were already scope-correct; only the seeder wasn't, so an officer opened their
  new site to three blank pages with no error. The registry is now scope-aware, seeds the
  three `park_*` blocks (including `park_meeting`, previously seeded nowhere), drops the
  parks page, and runs its copy through `OrgUnitNoun()`.
- **`park_meeting` printed a literal `0` as the venue name** (medium). `alternate_location`
  is a tinyint flag read as a name. The first fix made it *worse* — reading `location`
  instead dumped a raw geocode JSON blob on the page, because that column is a geocode
  cache in practice (522 of 806 rows hold JSON, 33 hold anything human). Caught by
  re-testing in the browser rather than assuming.
- **Seeded pages repeated their own title** — `Site_shell` already promotes the title to
  the page `<h1>`, so the leading heading block rendered it twice.
- **Empty `rich_text` / `cta_band` painted blank full-width bands** — now silent for
  visitors, italic author hint in preview, matching the existing hero/mosaic pattern.
- **`Route=Cms/edit` with no id returned a zero-byte HTTP 200** — blank page, no error,
  same shape as the DEV document-root bug fixed earlier on this branch. All four bare
  admin routes now redirect to their list, carrying scope.
- **TinyMCE logged an evaluation-mode warning** once per rich-text block — added
  `license_key: 'gpl'`.

`db-migrations/2026-08-10-cms-park-seed-repair.php` repairs park sites already seeded
with the kingdom template. Conservative by design: it retypes `kingdom_officers`/
`kingdom_events` to their `park_*` counterparts in place, **disables rather than deletes**
the two blocks with no park counterpart, and rewrites seed copy only where it still
matches the seed text byte-for-byte. It creates no pages, touches no nav, deletes
nothing, and is re-run safe.

`tests/cms-site` gains 13 checks pinning the scope-aware seeder. All five CMS suites pass.

---

## Update — the park starter template ("Seal and Field")

A park is not a small kingdom. It is a physical place a nervous stranger might drive to
on a Saturday, and its site has one job the kingdom's does not: convert that stranger
into someone who shows up. Parks previously got a lightly-trimmed copy of the kingdom
template. This replaces it with a purpose-built three-page site.

Design: `docs/superpowers/specs/2026-08-10-park-site-starter-template-design.md` ·
Plan: `docs/superpowers/plans/2026-08-10-park-site-starter-template.md`

### Two measurements decided most of the structure

Taken against a production-shaped restore of 342 active parks:

| Signal | Coverage | Consequence |
|---|---|---|
| Park-day records | **308 (90%)** | The recurring meeting is the spine of the site |
| **Upcoming event** | **26 (7.6%)** | **No Events page** — it would be empty on 92% of parks |
| Heraldry device | **316 (92%)** | The hero is a crest… |
| **Banner photo** | **5 (1.5%)** | …because it cannot be a photograph |

An Events *page* would have told a prospective newcomer *this club is dead* — via the nav,
before anyone clicked. Events stay as a block on Home, where the honest empty state reads
as "nothing beyond our regular park days."

### What ships

- **Three pages** — Home / New Players / Contact, against the kingdom's five. **About Us**
  is gone (its seeded body published author instructions to the open web) and so are
  **Documents & Resources** (a park has no library) and the **Board of Directors** roster
  (parks have no board; it published a fabricated person).
- **A sticky quick-facts strip** on every page — day, city, directions — sourced from
  park-day data, never hand-typed. All three audiences want that fact first and diverge
  only on the second, so it is hoisted above the split. Three degradation tiers: time and
  place; place only when there are no park days (**never an approximated time**); and
  nothing at all when neither exists.
- **`park_hero`**, a crest-led hero. The park's device is framed as a seal — `.jpg` devices
  cover-crop so their own background becomes the plate, `.png` devices float matted. The
  frame is the design decision; the image is cargo. Parks with no device get a monogram,
  not a shared placeholder crest. Carries the next game day and, inside a 7-day horizon,
  the forecast high.
- **A palette derived from each park's own arms**, extracted once at seed time and written
  to `ork_cms_theme`, so no two of the 342 look alike. Lightness is forced to 0.22
  specifically because white then clears 7:1 on **every** hue — worst case 7.11:1 at hue
  60, pinned by an all-hue sweep in the tests.
- **Copy that is true for any park unedited** — no weekday, no price, no city, no
  attendance claim, and never text that reads as instructions to the officer.

### Bugs fixed along the way

- **Vacant officer seats rendered as nameless cards** — the skip required *both* persona
  and role to be empty; a vacancy has a role. **187 of 342 parks** showed at least one.
- **`MedievalSharp` was the default heading font** and the seeder never created a theme
  row, so every org site shipped in a faux-medieval face.
- **Per-org theming never reached below the hero** — section bands and ~30 colours were
  hard-coded literals, so a wine-red park still got Bootstrap-blue links.
- **The member bar was hidden on phones** — right for the global front door, backwards for
  a park, whose most common visit is a member checking the time in a parking lot.
- **Four WCAG AA failures**, all found by measuring rendered pixels rather than reading
  markup: white-on-gold at **1.86:1** on the hero's primary CTA, the dark-mode accent
  token at 2.77:1, a kicker at **1.657:1**, and the monogram seal at 3.20:1 in dark. Worst
  ratio anywhere on the seeded pages is now 4.68:1.

### Two criticals only the whole-diff review could see

Every task passed its own review first; these needed the whole picture.

- **The heraldry colour extractor was dead for every kingdom.** `_heraldryPath()` built
  `%05d` for both scopes, but kingdom files are pad-4 (`0007.jpg`). The probe never
  matched, so kingdoms silently fell through to a name-hash colour — and the
  park→parent-kingdom cascade could never fire. The tell was in data already written: the
  stored primaries were *byte-identical* to the hash fallback. Pad widths now come from
  one `Heraldry::PAD_LENGTHS` constant, with a regression test.
- **`.fd-kicker`'s font-size was invalid CSS.** `--fd-font-scale` is a length (`1rem`), so
  `calc(var(--fd-font-scale, 1) * 0.6875rem)` is `rem × rem` — dropped silently on every
  front-door page. Measured at 14.67px (inherited) before, 11px after.

Also fixed: the hero and strip could display a park day **already in the past** (11 parks);
`park_hero` had no block-editor schema and so was uneditable; and the closing CTA band had
no "come to a park day" button — a URL-less park rendered zero buttons.

### Honesty about seeded content

`_parkIntroBody()` inherits the park's own `ork_park.description`, but **123 of the 246
parks with a description contain a weekday or a clock time**. Snapshotting one would
publish a meeting time that never refreshes — the exact failure the "never seed a time or
place" invariant exists to prevent, since a stale time sends a newcomer to an empty field.
Those descriptions now fall through to the evergreen paragraph.

### Deploy note

`db-migrations/2026-08-11-park-theme-backfill.php` must run **after** this lands. Any org
site seeded earlier carries a hash colour, and because the migration skips scopes that
already have a theme row, re-running it will not self-correct — those rows must be deleted
first.

### Not delivered

The design called for a clearly-labelled placeholder hero photo behind the crest.
`park_hero.tpl` supports `placeholder_image`, but nothing seeds it and it is unreachable
from the editor, so parks ship crest-and-colour only. Recorded in the spec's AS-SHIPPED
note rather than quietly dropped.

23 commits. All seven test suites pass (+27 new checks).

---

## Update — the CMS/CRM CSS separation, and the gates that keep it

The CMS ("OGRE") is a separate product hosted inside the ORK, but its CSS was
entangled with the CRM's. A standalone org site was downloading the whole ORK
application stylesheet to render a page that shares none of its chrome.

**Three surface tiers, enforced rather than intended.** A standalone org site at
`/k/{slug}` now loads `cms-base.css` + the public set + `orgsite.css` — and
**zero bytes** of `orkui.css`, `tokens.css` or `cms-admin.css`. In-shell CMS
surfaces keep the CRM pair plus `orkshell-interop.css`, the one designated
coupling point. OGRE admin renders inside the shell so it may *read* `--ork-*`
(350 times) but never define one.

`bin/check-css-boundaries.sh` holds the line with rules C0–C7 over **147 files**.
The scope is *derived*, not listed: everything under `frontdoor/` is public tier
and everything under `cms/` is admin tier at any depth; CMS page templates are
found from the controllers carrying the `CmsScopeContext` trait; and a
`$TPL_ROOT` template that renders CMS chrome is classified by *resolving* its
includes rather than matching a literal path. That last one matters — as a
literal match it hit no include line in the repo, because every CMS template
writes `include $fdDir . 'render_blocks.tpl'`, so a new routed CMS surface got
zero rules and could link `orkui.css` at exit 0.

The gate was then hardened against nine proven evasions, including a staged
symlink (`git show :path` returns the link target, one short line that passes
everything), CSS identifier escapes, attribute-selector spellings of the same
coupling, and `<style>` tags built by string concatenation in PHP. CMS PHP and
JS are in scope too: a verifier put a stylesheet link and a `<style>` onto a
live org-site page from two directions — `frontdoor.js` via
`insertAdjacentHTML`, and `controller.Site.php` echoing markup — and both gates
returned exit 0. That blind set was 31 files.

**Static CSS is now pinned tree-wide at `C3_TOTAL_STATIC=0`**, so any static
declaration inside an interpolating `<style>` trips on landing, and a
duplication ratchet (`MAX_GROUPS_2PLUS = 25`, `MAX_GROUPS_ANY = 89`) fails in
*both* directions — it demands a re-pin when duplication falls, so an
improvement cannot be silently given back. All four gates run in CI, where a
*skipped* runtime assertion fails the job rather than passing quietly.

Also here: org sites stop reporting pageviews into ORK analytics and stop
shipping the ~1 MB ORK JS bundle; `blog.css` links only on the two surfaces that
emit blog markup (it was 6,811 bytes matching zero nodes everywhere else).

The accounting is honest in both directions: a standalone org site got **52,953 B
lighter**, and the in-shell front door got **37,654 B heavier**, because
`blocks.css` is deliberately unconditional — block presence is authored content,
so linking it by current content would un-style the next edit. Every figure in
`frontdoor/css/README.md` carries the command that produces it.

## Update — OGRE's admin becomes an editor, not a form dump

A UX pass over the CMS admin, driven by watching the tool get used.

The block editor opened as a wall of every block's form at once; it opens as a
list now, bodies building lazily on expand. Card icons are **picked**, not typed
as CSS class names. Links are **chosen from a page list** instead of pasted as
`index.php?Route=` strings — and Navigation stops showing a route where a page
name belongs. Page types are named for what the author gets, with a sentence
saying what each one makes. The live preview shows what you are typing rather
than your last save. Media cards stop leading with **Delete**.

The dashboard leads with unfinished work instead of a second welcome banner, and
a brand-new site's dashboard invites work rather than reporting four zeros. The
rail and masthead became one workspace that uses the window; light mode got a
ground so cards read as objects rather than floating.

Bugfixes alongside: block summaries printed raw config tokens; image fields
showed a broken glyph instead of saying "missing"; a missing image file was
treated as an emergency rather than news; and the live preview rendered author
content on the admin's own origin — the one with the session cookie.

## Update — three review passes: layers, polish, simplify

A layer-separation and CSS-quality review, then a six-lens quality panel, then a
reuse/simplification pass. Findings were adversarially verified before anything
was applied, which mattered: **47 of 205 were refuted and 85 downgraded**. Every
"cross-kingdom IDOR via the model layer" finding died on inspection — the model
is a deliberately thin membrane and the callers do scope, with explicit
`IDOR guard` comments and `_rowInScope` / `_requireOwned` at each one.

**Defects that were real:**

- The dashboard's **Publish/Unpublish site button threw a ReferenceError on every
  click** — its `'use strict'` IIFE read a bare `UIR` this one template never
  declared. Nine templates were each re-declaring it while the shell already
  emits `window.CMS_UIR`, with three different escaping policies between them.
- The **CmsAjax CSRF gate was conditional on `REQUEST_METHOD`**, so a mutation
  taking no parameters executed on a bare cross-site GET.
- **OGRE admin gates read grant rows only**, never the officer-authority bridge,
  locking real officers out of the dashboard and page list.
- The **post editor autosaved a published post straight to the public site** —
  the page editor's deliberate published-guard was never carried over.
- **Route dispatch collided with the render step.** `Route=Site/view` answered
  200 with an unpopulated template. Fixed at the framework: `Controller::render()`
  now carries the render body, and both per-controller workarounds are gone.
- The **canonical-host allowlist protected two meta tags** while the config minted
  every absolute URL from the raw client `Host` — including `index.php`'s
  `header("Location: " . HTTP_UI)`, an open redirect. The Host is validated before
  any config runs; unset, behaviour is byte-identical.
- `SessionToken`'s **fail-open guard was dead code**: `YapoMysql::DataSet()` returns
  unconditionally and PDO runs `ERRMODE_WARNING`, so the blip it claimed to absorb
  still logged everyone out. A mitigation that reads as shipped while doing nothing
  is worse than none.

**Work that pays for itself:** park pages ran `GetParkDetails`/`GetParkDays` three
times each with staggered TTLs (~12 queries where 4 suffice); bulk media delete ran
an un-indexable `REGEXP` scan of `ork_cms_block.fields_json` per id; the nav cache
probe had grown heavier than the fetch it guarded; and a sparsely-populated org was
*permanently* uncacheable, because the cache could not tell a failed build from a
legitimately empty one — and the park strip renders on every page of an org site.

**Two things the reviews caught in their own output.** A first park-memo attempt
used instance properties and never hit, because `APIModel` constructs a fresh
domain object per call site — it was reverted, not shipped. And the batch
where-used API was built but never wired, so the perf win it advertised did not
exist until the follow-up pass connected it and deleted the public parameter that
could disable the safety guard.

70 commits. Layering, CSS boundary, stylelint and the duplication ratchet all
pass; 10 CMS test suites pass. Verified live in-browser: the 404 routes return
real bodies, the host guard rewrites an attacker `Host` when configured, the
preview renders `<body class="cms-page">` with the navbar gone, and org sites
serve no CRM CSS.

### Type across the CMS

The theme editor offered six families from a flat allowlist and drew their names
in the UI's own font, so choosing type meant guessing. It now offers **47** and
renders every name in the face it names.

`CmsThemeTokens::FontCatalog()` carries group, role, CSS fallback and the css2
weight axis per family, and the pickers, `Validate()`, `FontStack()` and the
public `<link>` all read it. Those were four hand-maintained lists — an
allowlist, an if-ladder of fallbacks, and hardcoded `<link>` tags in
`default.theme` — so adding a face took three edits in three files and
forgetting one failed **silently**. It already had: the org-site seeder wrote
Lexend for every site while `default.theme` never linked it, so every org site
asked for a webfont that was never loaded and fell back to the generic sans.

**The roles are a guard, not decoration.** Heading offers all 47; body offers the
21 faces readable at paragraph length, and `Validate()` enforces the same split
the picker shows — a hand-posted token cannot set body copy in blackletter. The
officer who did it would not see the damage from a one-line preview.

The picker is a real listbox because a native one **cannot do this**:
`font-family` on an `<option>` is honoured by Firefox and ignored by Chrome on
macOS. The `<select>` stays as the value holder and the no-JS fallback, and the
upgrade dispatches a bubbling `change` on it so the existing delegated handler
drives preview and dirty state unchanged. The panel is `position: fixed` and
placed from the button's client rect — `.te-group` is `overflow: hidden`, the
same trap the player-search autocomplete hit inside modals, solved the same way.

Loading is proportionate: picker rows fetch their own face as they scroll into
view, and a public page requests only the two families it renders, resolved from
its own theme. An org site went from **four hardcoded families to two
actually-used**, and MedievalSharp is no longer pushed to every org that never
chose it. Every family/weight combination was verified against the live css2 API
before landing — an unavailable weight makes Google answer 400 and the whole
stylesheet fails, taking the valid families with it.

**Two gates caught real defects here and neither was worked around.** C6 failed
closed on a stylesheet href built from a variable ("could name any file,
orkui.css included") — so the origin is now a literal in the template and only
the query interpolates. The duplication ratchet rejected a dark-mode block whose
every token `orkui.css` already flips; deleting it removed the duplication and
the redundancy together.

### Two bugs that only looking at the page could find

Both passed every automated gate.

**The park FAQ printed its own `<p>` tags at readers.** The starter template
seeded accordion answers as sanitized HTML into a field the rest of the pipeline
treats as plain text — the editor gives it a plain textarea, `accordion.tpl`
escapes it, and it is correctly absent from `CmsPage::HTML_FIELDS`. The seeder
was the one component that disagreed, and eight answers opened with a literal
`<p>`. Rendering the field raw would have been the wrong repair: nothing
sanitizes it, so un-escaping would trade a cosmetic bug for a stored-XSS
surface. Existing park sites need their stored rows repaired too.

**The live preview rendered a font nobody chose.** The preview is an `<iframe>` —
its own document with its own font set — so CSS naming a family the frame had
never fetched drew the generic fallback. That failure is quiet and *plausible*
rather than blank: `'Tangerine', cursive` with Tangerine missing renders in
macOS Snell Roundhand, so the preview showed a font, just never the picked one.

### Apostrophes in public URLs

`_normalizeSlug` hyphenated every non-alphanumeric, so "Angler's Rift" became
`angler-s-rift` — a bare "s" stranded between two words. **139 parks and one
kingdom** on this instance have an apostrophe in their name. An apostrophe sits
inside a word, so it is now deleted rather than hyphenated, and every spelling a
real name arrives in converges on one slug: ASCII, the curly quotes a word
processor or iOS substitutes, the modifier letter used in transliterated
orthographies, and the accents people reach for when they want an apostrophe. A
slug is a public URL and must not depend on which keyboard typed it.

Scope was checked rather than assumed: exactly **one** existing slug was
affected, so no back-fill is needed. That mattered — this helper also derives tag
slugs, and `_upsertTag()` uses the slug as its de-duplication key, so a
re-derivation that moved existing values could have split one tag across two
`ork_cms_tag` rows.

3 commits. Layering, CSS boundary, stylelint and the duplication ratchet all
pass; 10 CMS test suites pass, with 25 new assertions. Verified live on a park
org site in both themes.

---

## Code quality pass — make it read like it was authored

A formatting and documentation pass over the 176 CMS files this branch adds or
touches. No feature work and no behaviour change except one disclosed defect
fix. The goal was narrow: nobody reading this code should wonder whether a
person wrote it.

### Bookkeeping that leaked into the source

353 comments carried an internal review tag — `C13`, `E36`, `MOD-18`, `#121`,
"Task 2.1" — that resolves nowhere in the repo, densest in `class.CmsPage.php`
(51) and `controller.Site.php` (16). The prose after each tag was almost always
worth keeping, so the tags are gone and the sentences stay.

Two references survive on purpose: `check-css-boundaries.sh` really does print
rule IDs C1 through C8, so the two comments naming them point at something a
reader can find. That collision is also *why* the rest had to go — `C3` meant a
review ticket in `class.CmsPage.php` and a CSS gate rule in the gate script, in
the same tree.

### Comments that were wrong

`controller.Site.php` said breadcrumbs were "Dropped before this change"
directly above the line that builds them. `cms-admin.js` announced "the three
helpers" over a list of ten. `Site_shell.tpl` sent readers to
`cms-block-editor.js:1231` for the Level control, which is at 1719 — stale by
490 lines, with two more beside it wrong the same way, so cross-file citations
are now names instead of line numbers. A test claimed its predicate was
"extracted verbatim" from a template whose field names it did not match, so it
would not have caught a rename in the code it existed to guard.

Separately, comments that narrate history are rewritten to describe behaviour.
"Both USED to hand-call", "used to sit below the stat tiles" — git holds that,
and a reader who never saw the old code gets nothing from the comparison. The
reason each thing is the way it is survives; only the before-and-after framing
is dropped.

### One job, done several ways

`esc()` was hand-rolled four times in two dialects, in files that already
aliased `CmsAdmin.toast`. The confirm dialog existed as `cmsConfirmModal` copied
verbatim four times plus three private variants. Four list pages formatted the
same timestamp four ways, so Pages read "February 14, 2026" beside Posts reading
"Feb 14, 2026". Those collapse onto the `CmsAdmin` export and two new partials
under `cms/`.

Thirteen migrations opened with two incompatible CLI preambles and now share
`_cms_cli_bootstrap.php`, whose `UIR` is deliberately relative — seeds persist
those hrefs verbatim, so an absolute one bakes the seeding machine's origin into
content served everywhere. `park-theme-backfill` reached
`CmsSite::_seedOrgTheme` through `ReflectionMethod`; that method now has a
public entry point.

Dead code is gone, each confirmed by grep before deleting: sixteen model
wrappers tagged `@no-callers` that had none, two CSS classes, and two `UIR`
defines baking an absolute dev host a sibling migration explicitly forbids
baking.

### One real defect

`CreatePage` and `CreatePost` wrote their audit rows with action `'update'`,
under a comment describing them as "the content-creating write". They now write
`'create'`. The column is `varchar(40)` and `ork_cms_audit` has no reader today,
so nothing breaks — but the trail is append-only with no backfill, so a query
for `action='update'` splits at this commit. Flagging it rather than burying it.

### Deliberately not done

Porting `check-css-boundaries.sh` out of shell, moving the CMS suites onto
PHPUnit, and hoisting `UpdatePage`/`UpdatePost` into `CmsBase` are all
architectural rewrites. Doing them here would put the gates and the core write
paths at risk in the one pass that exists to avoid that.

Routing `Kingdomnew_index.tpl` through `Map::GetPublicParkMapLocations` was
written and then reverted. The dedup was right, but the lib pre-escapes park
names and `revised.js` escapes them again, so "The River's End" would have
reached readers as literal `The River&#039;s End` across **47 active parks**,
plus `stripslashes()` loss on 16 more. Fixing it properly needs `revised.js`,
which is outside this pass.

One finding is left open for its own pass: `htmlspecialchars($s, ENT_QUOTES)`
drops `ENT_SUBSTITUTE`, so invalid UTF-8 returns `''` instead of a replacement
character. That is the house idiom in **229 places** codebase-wide, not
something this branch introduced.

### Verification

Ticket codes went 353 to 2. PSR-12, PHP and JS syntax are clean; the CSS
boundary gate and stylelint pass; all nine runnable CMS suites are green
(`site_test.php` needed its assertion re-scoped to `_finishSeed`'s body, since
the new public wrapper is a legitimate second caller). The duplication ratchet
was re-pinned 89 to 88 because consolidating the member bar's inline styles
removed a duplicate group.

The layering gate reports three pre-existing violations that arrived with the
master merge — `default.theme:416`, `Playernew_index.tpl:99` and
`class.Controller.php:107`. All three fail identically on the commit before this
one, and this commit touches none of those call sites; the gate re-scans
`default.theme` only because the pass edited comments elsewhere in it.

The authenticated OGRE admin pages were verified statically only — the login
bypass is not active on this checkout, so the consolidations were confirmed by
cross-checking every call site rather than by loading the pages.

## Update — the media picker learns two views, and the teaser learns geography

Three commits: `f150afe5`, `bee211fc`, `956beb99`.

### The media picker was doing two jobs at once

Choosing an image to place is a fast, visual, constant task. Maintaining the
descriptions that make the public site accessible is a slow, textual, rare one.
Both were merged onto every tile, and the picking job lost: the author scrolled
past a drop zone, an alt field, a decorative tick and a four-line explainer
before reaching the first image, then met a grid where every tile carried its
own permanent input, Save button and checkbox.

Worse, the two alt-text fields — the one at the top writing to *the next upload*,
the one on each tile writing to *that media row* — were identically styled,
200px apart, with nothing distinguishing them.

A toggle in the modal head now names the two jobs, and the choice is remembered
per browser:

- **Grid** recognises an image by sight. Thumbnails keep their true aspect ratio,
  letterboxed on a new `--cms-field` ground, with filename, dimensions and weight
  beneath. Previously every image was cropped into the same 120×92 box by
  `object-fit: cover`, so a 6276×2437 banner and a 150×194 icon were drawn as
  identical rectangles — destroying the one fact an author most needs. Editing a
  description opens on one tile via a pencil rather than sitting on all two dozen.
- **List** audits and fixes. Every description is visible and editable in place,
  and any column sorts.

Upload moved into a collapsed panel titled "New upload", which is what finally
makes the two description fields tell themselves apart.

### Sorting is server-side, and that is the load-bearing part

`sort`, `dir` and `no_alt` go to `medialist` → `CmsMedia::ListMedia`, so each is
true of the **whole library** rather than of the rows that happen to have
lazy-loaded. Sorting by weight finds the heaviest image in the library, not the
heaviest of the first 24 — which is the entire reason to offer the sort.

`ORDER BY` takes no bound parameters, so the sort key is never interpolated: it
only ever *looks up* a fixed expression in `CmsMedia::$SORTABLE`, and an
unrecognized key falls back to newest-first. `CmsAjax::medialist` forwards
`$_GET['sort']` unvalidated **on purpose** — the lib whitelists against its own
table, so there is one authority for what is sortable instead of two that drift.
`CmsMediaSortWhitelistTest` pins it with 35 cases; it was mutation-checked by
bypassing the whitelist and confirming the test fails.

Every clause ends in a `media_id` tiebreak, so a sort survives the lazy-load
boundary without duplicating or skipping rows — verified across a page boundary
in the browser. `ListMedia` and the new `CountMedia` share one `WHERE` builder,
so the header's counts cannot drift from the rows beneath them.

### Fixes found along the way

- `_mediaListRow` **dropped `width`/`height`/`bytes`/`mime`**. The `SELECT` had
  always fetched them and the mapper always discarded them — which is why the
  picker could never show a dimension, or warn before a 3 MB PNG went into a page.
- The drop zone **promised 8 MB while `doUpload()` refused at 5 MB**. The copy now
  states the enforced limit, and nginx's `client_max_body_size` is raised to `8M`
  so the server is not the narrower bound.
- Each tile was a `role="button"` **containing a text input, a button and a
  checkbox**. Screen readers flatten a button's contents into its accessible name,
  so the inline description editor was unreachable by exactly the users it exists
  for. The tile is now a plain container with a real button around image and name.
- A **missing file showed an unlabelled grey square**, indistinguishable from one
  still loading. It now uses the admin's existing `.cms-missing-thumb` treatment —
  the same one the media grid and the block editor's image fields already use —
  rather than the third private implementation this work first grew.
- A **stale rule block** later in `cms-admin.css` was overriding the picker's grid
  columns, editor layout and Load-more margin. Removed.
- `.cms-media-modal` also carried `.cms-modal-wide`, which is more specific, so
  **the modal never reached its own width**.

### Kingdoms Teaser: optional states/provinces

An opt-in toggle adds a line to each kingdom card naming the distinct
states/provinces its active parks sit in, so a visitor can find the realm nearest
them without opening every profile. `Report::GetActiveKingdomProvinces` resolves
every kingdom in **one** grouped scan, not a query per card, and folds a
principality's parks into its parent kingdom.

`ork_park.province` is free text, so `NormalizeProvinceList` expands US/Canadian
postal codes (collapsing "OH" and "Ohio" into one entry), dedupes
case-insensitively and drops blanks; `ReportProvinceListTest` pins it. The
Freeholds is landless with parks across dozens of states, so it gets a fixed
label instead of a list.

Two supporting fixes: `CmsRenderCache::TeaserKey` now carries the toggle, or the
two renderings share a cache entry and whichever rendered first wins for
everyone. And the block editor's `bool` field gained a `default` — an absent bool
read as **Yes**, so adding this field to an existing block type would have
switched it on for every already-saved block merely by opening it.

### Front-door parks map

The lazy-load `IntersectionObserver` watched `[data-kpm-map]`, which is
`display: none` until `init()` runs. A `display: none` element never reports
`isIntersecting`, so the observer meant to trigger `init()` could not fire — the
map waited on a callback waiting on the map. It now observes `ROOT`.

The block also had no Maps key outside production and rendered "map unavailable"
everywhere else. **It now falls back to the ORK's shared browser key, which
reverses an earlier hardening** — so it is isolated in its own commit
(`bee211fc`) and easy to revert. The same key is already committed and already
served in public page source by `Kingdom_map.tpl`, `Atlas_index.tpl`,
`Atlas_map.tpl` and `revised.js`, so this adds no exposure that does not already
exist — but it is a fifth place to change if the key is ever rotated, and
referrer-restricting it in the Google console is what actually bounds the
quota-abuse risk. Worth a reviewer's eye.

### Deliberately not done

The List view is **not** a DataTable, against the house rule for tabular data.
Its rows are form controls (the documented exemption), and DataTables' own
search and paging would have fought the server-side sort and the existing
lazy-load. Easy to revisit if the exemption is read more narrowly.

### Verification

456 tests pass. The layering and CSS-boundary gates are clean. The duplication
gate caught ten new groups from this CSS; all were collapsed by selector
grouping — including reworking reveal-on-hover into `@media (hover: hover)`, so
those controls are not unreachable on a touch screen — which left the file one
group below the old floor, so `MAX_GROUPS_ANY` is pinned **down** 88 → 87.

Both views were driven in the real app in Chrome, light and dark: view toggle and
its persistence, search-as-you-type, the "needs a description" filter, sorting
each column, editing a description from both views (confirmed written through to
the database, then restored), picking from both views, Escape clearing the search
before closing the modal, the sticky table header, and sort holding across a
lazy-load page boundary. Four candidate designs were staged and reviewed before
this pair was chosen.


baltinerdist:

Synced up to `master` @ `6ab028a3` (merge commit `710ad021`). This branch was **342 commits behind**; now 0 behind / 139 ahead. Only 6 files conflicted — the rest of the gap auto-merged.

## Conflicts and resolutions

**`system/lib/system/class.Controller.php`** — the substantive one. Upstream did an architecture-layer refactor (raw `$DB` out of the controller, into models) while this branch split `index()` into `_indexCommonData()` + `_indexFrontDoor()` and added CMS/brand plumbing. **Both integrated**: kept the branch's `PUBLIC_SITE_BRAND`/`APP_BRAND` consts, `_csrfToken()`, `CanManageCms` probe, the `index()` split and the whole `_indexFrontDoor()` payload — then applied every upstream change on top:

- `Controller_EventEmbed` added to the token-check skip list (union with the branch's `Controller_CmsAjax`)
- session token check → `Model_SessionToken::validate_session_token()`
- viewer font prefs → `Model_Player::get_viewer_preferences()`
- **upstream's `ShowWhatsNew`/`WhatsNewRelease` computation added** — mandatory, not optional: the merged `default.theme` reads those keys and this branch never set them, so omitting it would have broken the What's New modal
- all three `HasAuthority` calls → `$this->Authorization->has_authority`, plus `$this->data['NavIsOrkAdmin']`
- constructor `$DB->Clear()` → `$this->Authorization->clear_db_after_auth_checks()`
- home-kingdom lookup → `Model_Player::get_home_kingdom()`; RSVP counts → `Model_Event::get_rsvp_total_counts_batch()`

**`orkui/template/default/default.theme`** (3 hunks) — (1) Help-panel CSS: upstream commit `a3236154` deliberately deleted the unfinished Help button (never wired to a backend, always showed "No help topics available"); took upstream's deletion and confirmed zero remaining `nav-help-btn`/`ork-help-panel` references. (2) Nav dropdown: kept the branch's "OGRE Site Manager" link but switched its gate to upstream's precomputed `$NavIsOrkAdmin`. (3) Kept the branch's `if (empty($IsOrgSite))` guard around the What's-New chrome — org sites lack the base-Controller data and would fatal mid-render.

**`orkui/template/revised-frontend/Kingdomnew_index.tpl`** (5 hunks) — banner auth moved to the controller upstream (`knCanManageBanner`, the identical `AUTH_KINGDOM/AUTH_EDIT` gate this branch computed inline) → took upstream's and verified the controller sets it. Admin tab/panel: kept the branch's ARIA wiring (`role="tab"`, `aria-controls`, `role="tabpanel"`, `aria-labelledby`) **and** upstream's widened gate `|| !empty($CanManageTests)` for qual-test reeves. Modal headings: kept the branch's `id=` (its `aria-labelledby` needs them) **and** upstream's icon renames — upstream's `5c7e1db8` swapped `fa-people-arrows`→`fa-exchange-alt` and `fa-compress-alt`→`fa-compress-arrows-alt` because the originals are FA 5.13+/FA6 and render blank on this repo's FA 5.8.2.

**`orkui/model/model.Kingdom.php`** (4 hunks) and **`system/lib/ork3/class.Kingdom.php`** (4 hunks) — unions; no duplicate method names. From upstream: `QualTest*Enabled` seed configs, the hardened `GetParks` failure response (`array('Status' => …, 'Parks' => array())` so `Kingdom/map` never `array_filter(null)`), and the post-config memcache flush. **`.gitignore`** — additive both sides, kept both.

**Verification:** no conflict markers; `php -l` clean on all 287 changed `.php`/`.tpl`/`.theme`; `class.Authorization.php` byte-identical on both sides (blob `248b8d29`), no bypass. Front-door arrangement independently re-confirmed on the pushed branch: `Controller::index()` → `_indexCommonData()` + `_indexFrontDoor()` (L215–218), `Controller_Directory::index()` → `_indexCommonData()` only.

---

**Three things for a human:**

1. ⚠️ **One non-conflicted file was edited: `system/lib/ork3/class.SessionToken.php`.** This branch shipped an explicit stability fix (commit `7c0792df`: session-token check no longer logs users out on a transient DB error) by distinguishing `$_rs === false` from a real token mismatch. Upstream's refactor moved that check into `SessionToken::ValidateSessionToken()`, whose `if (!$rs || !$rs->Next()) return false;` returns "invalid" on a DB blip — silently reverting the fix. The semantics were ported into the new home (a 4-line `if ($rs === false) { return true; }` with a comment) rather than dropping the fix or refusing upstream's structure. **If you'd rather upstream's semantics win, revert that hunk.**
2. **`GetHomeKingdom()` keeps a `Size() > 0` guard** that this branch had deliberately removed as "unreliable"; upstream's lib version reinstates it. Upstream's code was left alone (buffered `rowCount()` is correct for these SELECTs), but if the author had a concrete failure in mind, `$this->data['UserKingdomId']` could come back 0.
3. **Layering:** the branch's `ViewerName` block in `_indexCommonData()` still does a raw `$DB->DataSet()` in the controller. It has no upstream model equivalent, so moving it would have been a drive-by during a merge — but it now sits conspicuously next to the model-based reads and is a candidate for the ongoing domain migration.

baltinerdist:
## Seeding a local (or any other) environment

Short version: **`db-migrations/seed-cms-demo.sh` is the one command.** It is idempotent — safe to re-run — and it now seeds the amtgard.com replication from specs committed to the repo rather than skipping it.

```bash
# from the repo root, with the dev containers up
db-migrations/seed-cms-demo.sh
```

That gives you the schema, the front-door home page, the exemplar and staff-roster pages, all 15 amtgard.com replication pages, the marketing nav with every relink, and an exemplar blog post.

If your container/DB names differ from the compose defaults:

```bash
APP_CONTAINER=ork3-php8-app DB_CONTAINER=ork3-php8-db \
DB_USER=ork DB_PASS=secret DB_NAME=ork db-migrations/seed-cms-demo.sh
```

### Why this needed fixing

The 15 replicated amtgard.com pages could not be reproduced from a clean checkout. The seed reads a staging directory produced by scraping the live site; that directory is gitignored (`db-migrations/.amtgard-assets/`) and the original was a session scratchpad that no longer exists. Nothing was ever committed — `git rev-list --all` finds zero objects under that path on any branch.

The practical effect: a machine that had already run the seed kept the **result** in its database and lost the **inputs**. So the content looked fine there and was unreproducible everywhere else. On my checkout the pages were present and every one of their 62 images was already broken, because `assets/cms-media/` is gitignored too and had been cleaned up.

### What is committed now

`db-migrations/amtgard-specs/specs/*.json` — all 15 pages: structure, copy, hierarchy, meta descriptions, and the filename of every image and document each block references. 82 blocks.

**Asset bytes are deliberately not committed.** They are ~27 MB of scraped third-party binaries at full resolution and do not belong in git history. Two scripts cover the round trip instead:

| Script | Direction | Use it when |
|---|---|---|
| `db-migrations/fetch-amtgard-assets.php` | live site → staging assets | you want the imagery |
| `db-migrations/export-amtgard-specs.php` | seeded DB → specs | you changed pages in the CMS and want that captured in the repo |

### Getting the imagery

```bash
php db-migrations/fetch-amtgard-assets.php db-migrations/amtgard-specs
db-migrations/seed-cms-demo.sh          # re-run; it picks the assets up
```

**Be aware this is partial today, and the script will tell you which parts.** Each spec names its assets by slot and extension (`1.jpg`, `2.png`, `3.jpg`). The fetcher walks the live page's content images in DOM order and pairs them off, but only when that extension sequence still matches. Where amtgard.com has drifted since the July scrape it prints `IMAGE MISMATCH` and writes nothing for that page — a silently wrong image is worse than a missing one.

As of this comment it recovers **14 assets across five pages** (about, mission, programs, learn-the-basics, start-a-chapter). Two known gaps:

- **galleries** (24 images) — the Wix Pro Gallery loads through an API call, so the image URLs are not in the page HTML at all.
- **documents / programs PDFs** (15 docs) — served under hashed `ugd/<hash>.pdf` URLs with empty link text, so there is nothing to match the friendly filenames against.

Both need a re-extract of those pages rather than a smarter fetcher. Everything else — all the page copy, the staff roster, the hierarchy, the nav wiring — seeds completely without them.

### If you already have a seeded database

Re-running the seed is safe and now repairs something it previously could not. The media dedup used to reuse any `ork_cms_media` row with the expected filename **without checking the file still existed**. Because `assets/cms-media/` is gitignored, a database seeded on one machine and used on another names files that are not there — and the seed handed those rows back forever, reporting success on every run while the pages rendered broken images. It now verifies the bytes are on disk and reclaims the row if they are not, so a re-run after `fetch-amtgard-assets.php` actually heals the page.

You can confirm what your environment has:

```bash
# live replication pages (expect 15)
docker exec ork3-php8-db mariadb -uork -psecret ork -N -B -e "
SELECT COUNT(*) FROM ork_cms_page WHERE scope_type='global' AND deleted_at IS NULL
AND slug IN ('about','join','programs','media','resources','mission','staff','volunteers',
'learn-the-basics','start-a-chapter','foodfight','olympiad','galleries','writing','documents');"

# media rows whose file is actually present
docker exec ork3-php8-db mariadb -uork -psecret ork -N -B -e "
SELECT path FROM ork_cms_media WHERE filename LIKE 'amtg-%' AND deleted_at IS NULL;" \
| while read -r p; do [ -f "assets/$p" ] && echo ok; done | wc -l
```

### If you have the original staging directory

If anyone still has a real `.amtgard-assets/` from the July extract, it wins — `seed-cms-demo.sh` prefers it over the committed specs when present, so nothing here overrides a complete local copy. If you do have one, dropping it in and running `export-amtgard-specs.php` afterwards would let us commit the last of the asset filenames properly and close the two gaps above.

PR #490, opened 2026-07-11
Enhancement: Court Planner + Recommendations workflow (Planner, Court Report, Herald Script, Recs Manager, notifications, pass-to-local)

baltinerdist:
## Court Planner + Recommendations workflow

A complete workflow for planning royal courts, running them, printing the paper packet a herald actually carries, recording what happened afterward, and managing award recommendations.

### Court Planner
- Plan kingdom/park courts: schedule, add awards + contributing artisans, drag-reorder, scroll/regalia tracking, and grant directly from the plan
- **Stage → finalize**: granting at court captures giver/reason/date/rank and marks the line `staged`; it does **not** touch the permanent record. A single **Finalize** (folded into Complete Court) batch-commits every staged grant, so undo before finalize is free. The idempotency key is the court line — an atomic `staged→given` claim means a line commits at most once, and finalize is throw-safe (a failed `AddAward` reverts that row to `staged` and the court will not complete until it is clear)
- **Run vs Plan** is a stored court mode. Run mode shows a granted row as Given with an Undo for the one-tap ceremony flow; Plan mode keeps the explicit staged framing. Both share the identical pipeline
- **Optimistic concurrency** (`row_version`) plus a presence heartbeat, so several reeves can run one court live without silently clobbering each other
- Editable court details — name, date, linked event, and **recorder** — after publication
- **Court Report** (public): per-kingdom/park court history with awards given
- Herald script generation for reading court
- Responsive/mobile layout for running court from a phone

### Court packet (print)
The printed output is a three-sheet packet sharing one page geometry, chosen from a sheet selector on the court:
- **Sheet 1 — Order of Court:** the running order the herald reads from
- **Sheet 2 — Court Record:** the write-on sheet, with the columns the Record Court screen mirrors exactly, so paper reads back in one-to-one
- **Sheet 3 — Prep Sheet:** grouped by maker, for scroll and regalia prep

Printing stamps the court (`last_printed_at` plus the award count at print time), and the court warns when the plan has **drifted from the printed packet** — measured by award count, not a modified timestamp, so a cosmetic edit does not cry wolf. Skipped awards print struck through rather than silently vanishing, so the paper matches the plan the herald walked in with.

### Record Court
A dedicated catch-up view at `Court/record/{id}` for typing the marked-up paper back in after the ceremony:
- Rows mirror Sheet 2's columns; marks are **eager server writes** through the endpoints that already exist (`grant_award` / `skip_award` / `unstage_award`) — no new grant path, and Finalize & Complete stays the single commit
- Per-row marks plus **mark-all-remaining**, a giver chip with **apply-to-rest**, and inline rank selection
- Inline **citation editing with partial save**, showing an inherited citation rather than a false "No citation", and correctly distinguishing an untouched inherited box from the recorder's own text
- Inline **walk-on row** for awards given at court that were never on the plan
- Built on the standard `rp-*` tool-page shell with a densified award grid; shared `cp-*` chrome extracted into `court-planner.css`
- **Recorder assignment** (the Prime Minister by default) with an in-app notification for courts that were published but never recorded, and a surfaced list of those unrecorded courts

### Recommendations Manager
- Dedicated management page, **server-side paginated** (500-row batches + infinite scroll), grouping parallel recommendations into one row per (recipient, award, rank) cluster
- Server-side filter/sort; **Park** and **Rank** columns; **CSV export** of the full filtered set
- **Grant Award modal** — pre-filled from the recommendation, with rank pills (green = already held), officer quick-pick chips (Monarch/Regent), and a park/kingdom/event location picker
- Granting a recommendation that already sits on a court plan is reconciled **server-side** in the same request, scoped to courts the officer can manage and never to a completed court, so a later finalize cannot re-grant it. The officer's "leave on court" / "remove from court" choice is carried as `CourtAction` and applied in the domain layer
- Snooze-to-next-monarchy, pass-to-local delegation, and second/support a recommendation
- **Anonymous recommendations** (recommender masked per viewer, never cached across viewers)
- In-app **notifications** to advocates when a recommendation is granted
- Responsive **card collapse below 700px**, a 44px checkbox tap area (22px glyph), no tablet dead band over the Actions column, and a select-all that states plainly that it covers loaded rows only

### Architecture
DB work lives in `system/lib/ork3/` (`Court` + `Report` libs); controllers stay thin and reach the domain layer through `orkui/model/model.Court.php`.

### Migrations
Thirteen migrations under `db-migrations/` — court tables, stage/finalize + grant-safety columns, the recorder column, the printed stamp and printed award count, recommendation snooze / pass-to-local, the notification table, and an idempotent repair for `ork_session.user_agent`/`ip`. Apply before deploy.

> The session repair is unrelated to court but ships here because the branch surfaced it: the device-metadata columns were added by editing the original `2026-07-13-add-ork-session-table.sql` in place. That file leads with `CREATE TABLE IF NOT EXISTS`, so re-running it on a database built from the pre-metadata version adds nothing, `CreateSession()` then inserts columns the table does not have, and **every password login fails**. Fresh installs were never affected.

### Review

The branch has been through four review passes, each of which found real defects in the previous one's output. Everything found is fixed unless called out under **Known gaps** below.

**Adversarial panel (hostile, findings kept only if they survived an organized attempt to refute them).** Two rounds, then a third after a layering refactor. The panel dropped three of its own loudest findings as false positives, which is worth stating: the "critical IDOR" on dismiss/snooze/pass-to-local was wrong (all four enforce authority inside the lib, scoped tighter than the route), and two "anonymous recommender leaks" were wrong (`applyViewerFlags` already masks per request).

What it did find, and this branch fixes:
- **The same honor could reach `ork_awards` twice.** Every duplicate guard was per-request (`$committedClusters`) or per-court (`addAward`'s probe); `Player::AddAward` has none. A retried finalize or the same honor staged on two courts wrote two permanent rows silently. The guard took three attempts to get right — `rank >=` would have blocked 22 of 22 real lines on one production court, and bare rank equality still blocked 14 of 22 because it was time-blind (1,584 exact-rank repeats since 2022 are more than a year apart, and repeatable tournament titles like Weaponmaster carry `is_title=1`). Scoping to the court date blocks 0 of 22 while still catching a retried finalize by construction.
- **The public Court Report published ceremonies that had not happened.** It keyed on the award line's status and never read `court.status`, so a draft court with one line flipped to `given` published as a held ceremony — leaking a confidential court plan to a login-free page.
- **1,307 live recommendations for retired players were unreachable.** An `m.active = 1` predicate added earlier on the branch hid them from every officer surface with no filter to bring them back and no suppressed count. Retirement and memorial honors are precisely the ones given to people who have stopped playing.
- **"Snooze to Next Monarchy" tracked the wrong throne** — the recipient's park Monarch even for a kingdom-scope snooze, so a Coronation never lifted it while an unrelated park election did.
- Plus a badge/filter scope mismatch and a printed packet that used the pre-drag order.

**Distributed review** — a seven-lens panel (five technical, plus a herald and a kingdom Prime Minister reasoning about how courts actually work) raised 62 findings, synthesised to 50. 47 are fixed. The highlights beyond the above:
- The Recs Manager's Grant — the path officers use daily — had **no ledger guard at all**, and its post-write court reconcile ran unguarded, so a throw after `AddAward` already committed reported "Grant failed" and re-enabled Submit. The duplicate check is deliberately **advisory**: it never refuses, because a legitimate re-earning and a mistaken re-grant are indistinguishable to it, and refusing the correction path is worse than the duplicate.
- **Finalize published the recommendation's private reason as the public citation even after the officer clicked Clear.** That intent is now persisted and honored, and the flag round-trips through `getCourtAwards` and the heartbeat reconcile — without that, a peer's save posts `cleared=0` under a matching `row_version` and silently republishes confidential text.
- Per-kingdom `is_ladder`/`is_title` overrides are read on **both** axes via `GREATEST`, not `COALESCE`: those columns are `NOT NULL DEFAULT 0`, so `COALESCE` can never fall through, and the literal reading would have stripped the ladder flag from 607 kingdomawards.
- The printed packet emits all three sheets rather than stamping the whole packet after printing one — an officer was walking into court with the herald's sheet and no Court Record sheet, which is the entire input to Record Court.
- Dismiss counts recommendations rather than clusters (10 selected rows could retire 40 people's filings).
- The public report names who conferred each honor (it credited the scroll and regalia makers but not the giver) and stacks into cards below 600px, where its four-column table had been running 458px wide inside a 375px clip.
- Accessibility: row actions, both modals and the shared dialog are keyboard- and screen-reader-operable; one shared `orkConfirm`/`orkAlert` replaces three hand-rolled copies.
- A composite index backs the Manager's cluster paging; CSV export streams in batches.

### Known gaps

- **No amend path for a finalized court.** Only interim gates ship: `add_award` and `reorder_awards` now refuse writes to a complete court, which was silently reordering the public report for a ceremony already held. Correcting or voiding a finalized line needs a product decision about who may amend the record and whether voiding revokes the linked award.
- Three items from the distributed review are deferred with the above.
- Pre-existing and not introduced here: `.rm-btn-primary` is 2.33:1 in dark mode (every primary button, not just this feature).

### Deploy

Thirteen migrations under `db-migrations/`. Apply in date order; the three dated 2026-09-09 are new. Two pre-flight checks against production before deploying:

```sql
-- 1. These rows will DISAPPEAR from the public Court Report (given lines on courts
--    that were never completed). 0 on the prod-derived snapshot.
SELECT COUNT(*) FROM ork_court_award ca JOIN ork_court c ON c.court_id = ca.court_id
WHERE ca.status = 'given' AND c.status <> 'complete';

-- 2. A non-zero result blocks the new unique index in
--    2026-09-09-court-planner-review-fixes.sql. Dedup, then re-run the migration;
--    it is ordered so the ADD COLUMNs land first and only the index can abort.
SELECT COUNT(*) FROM (
  SELECT court_id, recommendations_id FROM ork_court_award
  WHERE recommendations_id IS NOT NULL AND recommendations_id <> 0
  GROUP BY court_id, recommendations_id HAVING COUNT(*) > 1) d;
```

> One migration is unrelated to court but ships here because the branch surfaced it: `ork_session`'s device-metadata columns were added by editing the original table-creation file in place. That file leads with `CREATE TABLE IF NOT EXISTS`, so re-running it on a database built from the pre-metadata version adds nothing, `CreateSession()` then inserts columns the table does not have, and **every password login fails**. Fresh installs were never affected.

### Verification

Integration **264 tests / 948 assertions** and unit **127 / 390** pass. `php -l` clean, pinned php-cs-fixer clean, 99 migrations classified. The Court Planner, Record Court, the public report and the Manager were each exercised in a browser, with dark-mode and contrast claims measured by computed style and mobile measured in a 390px same-origin iframe.

**Not covered:** no concurrency was exercised, so the multi-reeve and atomic-claim guarantees are reasoned from the SQL shape rather than reproduced under load; and the printed sheets were inspected as DOM rather than actually printed.

https://claude.ai/code/session_016XHnYr7jkzChUZkaaKgKHb


baltinerdist:
Updated with the Court Planner **grant-safety + workflow/UX** implementation (commit `748e3df6`):

**Records integrity**
- Single idempotent grant sink (`Court::commitStagedAward`) — court-line identity is the idempotency key; `Player::AddAward` now returns the real inserted id (date-heuristic removed); throw-safe finalize (revert-on-throw, re-runnable).
- Server-side cross-path reconcile so a Recs-Manager grant can't be re-granted at finalize; `update_court_status` rejects `complete`; remove/skip/setStatus guard `given` rows.

**Concurrency**
- `ork_court_award.row_version` optimistic lock (stale → non-destructive reload), full-field heartbeat reconcile, presence roster + honest sync indicator.

**Workflow / mobile / a11y**
- Run-mode "Grant" + auto-finalize on Complete; walk-on adds while published; inline un-skip; mobile stacked-card layout so Grant/Skip are reachable on a phone; a11y additive pass; native confirm/alert removed.

**Ceremony polish**
- WCAG-AA rank-pill ramp; functional status (Cancelled → Skipped); expanded-row highlight; flags/type tooltips; ad-hoc add split into **Add Award / Add Title** with typeable autocompletes grouped exactly like the player Add Award modal (rank pills replacing the number box).

Verified end-to-end against a live instance (6/6 functional + 4/4 fix re-tests; all data-integrity SQL invariants at 0). Migrations: `2026-07-11-court-stage-finalize.sql`, `2026-07-12-court-grant-safety.sql`.

baltinerdist:

Synced up to `master` @ `6ab028a3` (merge commit `8cb99dc5`). This branch was **263 commits behind**; it's now 0 behind / 130 ahead.

**Conflicts: 3 files, resolved.** All three were upstream's domain-migration work landing on top of Court Planner additions.

**1. `orkui/model/model.Award.php`** — upstream's R-10 migration gutted this model into a thin pass-through (`fetch_award_option_list()` → `Award::GetAwardOptionListHtml()`, plus an `_award()` helper). This branch had added a Court-Planner-only `fetch_award_option_groups()` for the ad-hoc Add Award/Title autocompletes, and otherwise only reformatted `fetch_award_option_list` (confirmed via `git diff -w base..HEAD` — no logic change on the branch side).

Resolved to upstream's thin version **plus** this branch's `fetch_award_option_groups()` and `compareAwardsByName()` kept verbatim. Note the deliberate choice: `fetch_award_option_groups()` was **not** rewritten on top of upstream's new `Award::GetAwardOptionGroups()`, because the two return different shapes — this branch returns ordered `[{label, options:[…]}]` with per-group `IsTitle` and Custom split into "Custom Award"/"Custom Title" groups, while upstream returns `Groups`/`StandaloneOptions` with raw award rows. Adapting one to the other mid-merge would have risked a silent behavior change in the Court picker.

**2. `orkui/model/model.Player.php`** — upstream PSR-12'd the file and added `reset_waivers()`/`add_player_recommendation()`; this branch added four Court/Recs methods in the same region. Kept upstream's structure and re-added all four (`resolve_player_recommendation_cluster`, `snooze_recommendation`, `unsnooze_recommendation`, `set_recommendation_passed_to_local`) in upstream's PSR-12 style. Verified no duplicate declarations, and that the corresponding lib methods on `class.Player.php` all survived the auto-merge.

**3. `orkui/controller/controller.KingdomAjax.php`** — upstream (R-18) replaced the inline `geteventtemplates` SQL with `$this->Event->get_event_templates_for_kingdom()`; this branch had inserted `snoozerecommendation`/`unsnoozerecommendation` handlers immediately above it. Kept upstream's model call **and** both branch handlers. `Event::GetEventTemplatesForKingdom()` was checked to return the identical shape (`EventId`/`Name`/`ParkId`/`ParkName`) the JS consumes, so no client-side breakage.

**Method:** `git merge upstream/master`, not a rebase — this branch has 130 commits behind prior merge commits, and a linear rebase would replay all of them against the new base.

**Verification:** no conflict markers; `php -l` clean on all 270 changed `.php`/`.tpl`; duplicate-method scan across 239 changed PHP files found none; `class.Authorization.php` untouched by both sides with no bypass. Feature spot-checks post-merge: the stage/finalize state machine in `class.Court.php` is intact (`'staged' -> 'given'` atomic claim + revert paths), and Recs Manager is still server-paginated (`'Limit' => 500` + `NextOffset`, infinite scroll in `Recommendations_manage.tpl`) — independently re-confirmed on the pushed branch.

---

**Worth a maintainer's eye:**

1. ⚠️ **The `snoozerecommendation` / `unsnoozerecommendation` handlers have no controller-level auth guard**, unlike the adjacent `restorerecommendation` block which returns `status 5`. That's how they shipped on this branch — presumably auth lives in the lib — and it was preserved as-is rather than hardened mid-merge. Worth confirming the lib really does gate them.
2. **Award classification logic is now duplicated** between `Model_Award::fetch_award_option_groups()` and upstream's `Award::GetAwardOptionGroups()`. A consolidation follow-up, out of scope for a merge.
3. **Unrelated upstream change to be aware of:** upstream's new `GetAwardOptionListHtml()` emits Custom options *after* the optgroups, where the pre-merge code emitted them between Ladder and the rest (upstream commit `6d06e4c0`). Not introduced here, but it changes picker ordering.
4. The two snooze handlers were re-indented from tabs to spaces and their one-line `if (!valid_id(...)) {…}` guards expanded, to match the now-PSR-12 file. No logic changed; no php-cs-fixer was run.

baltinerdist:
### Follow-up: snooze guards verified, Court model layer added

Two items from the sync comment above, resolved.

## 1. Snooze handlers are authorized — **no guard added, and none needed**

I asked for evidence before adding anything, and the evidence says the lib enforces it. `Model_Player::snooze_recommendation` / `unsnooze_recommendation` (`model.Player.php:235,240`) are bare pass-throughs to `class.Player.php`, which does a **two-stage** check.

`Player::SnoozeAwardRecommendation` (`class.Player.php:3954`):
```php
if (($mundane_id = Ork3::$Lib->authorization->IsAuthorized($request['Token'])) == 0) {
    return NoAuthorization();
}
...
// Auth: must be park admin for recipient's park
$recipientInfo = $this->player_info($awardRec->mundane_id);
if (!Ork3::$Lib->authorization->HasAuthority($mundane_id, AUTH_PARK, $recipientInfo['ParkId'], AUTH_EDIT)) {
    return NoAuthorization();
}
```
`UnsnoozeAwardRecommendation` (`:3999`) carries the identical pair.

**The important detail:** the scope is derived from the **loaded recommendation's recipient** (`$awardRec->mundane_id` → their `ParkId`), never from a request parameter — so it cannot be spoofed by passing someone else's `kingdom_id`. That makes it *stricter* than `restorerecommendation`'s controller-level `AUTH_KINGDOM/AUTH_CREATE` check, not weaker. No IDOR. Independently re-confirmed on the pushed branch.

Same applies to the identical unguarded pair at `controller.ParkAjax.php:427,439` — covered by the same lib enforcement.

## 2. Court model layer (`fb1c2888`)

**69 call sites converted** across 7 controllers — noticeably more than the ~44 estimated: `CourtAjax` 45, `Court` 11, `Kingdom`/`Park`/`Recommendations`/`Reports` 3 each, `PlayerAjax` 1. `grep -rn 'Lib->court->'` now returns **nothing repo-wide** (verified post-push).

New `orkui/model/model.Court.php`: 39 thin pass-through methods, **zero SQL, zero logic** (confirmed by grep). It follows `model.QualTest.php` — the closest analogue on this branch, also fronting a camelCase lib with snake_case wrappers and a private accessor. (The two models I'd suggested as templates, `model.RBACService.php` and `model.OfficerPosition.php`, don't exist on this branch — they're on #475.)

**Name resolution is auto-wired**, no registration needed: `Model::__construct` splits `Model_Court` on `_` and takes `Court`; `startup.php:52-56` has already `require_once`'d every `DIR_ORK3/class.*.php`, so `Court` is defined before any model instantiates.

Two decisions beyond the mechanical:
- **`load_model('Court')` placement** — constructor for the Court-centric controllers (`Court`, `CourtAjax`, `Recommendations`, where nearly every action needs it and `Recommendations` uses it inside the shared `resolveContext()`); per-action for `Kingdom`, `Park`, `PlayerAjax`, `Reports`, matching each file's local convention.
- **Param/return types deliberately omitted** on the model methods — the `Court` lib methods are untyped, and adding declarations would introduce coercion that could change behavior.

`controller.Reports.php` is tab-indented (60 tab lines) while the rest are spaces; edits there were surgical and tab-preserving rather than running php-cs-fixer over an otherwise-untouched file.

**Verification:** `php -l` clean on all 8 files; a static cross-check confirmed all 39 model methods forward to a real `Court` method with matching total and required arity, and all 69 call sites pass an argument count within range. A stub harness then instantiated `Model_Court` against a capturing `Court` stub and invoked all 39 public methods by reflection — **39 dispatched, 0 failures**, each forwarding the full parameter list including defaults.

**Not runtime-tested in the app** — the worktree isn't what the local Docker instance serves. Verification is static plus that isolated dispatch harness.

PR #491, opened 2026-07-13
Enhancement: Multi-device login — 3 concurrent sessions with LRU eviction

baltinerdist:
## Summary

Replaces ORK3's single-device login (one `ork_mundane.token` slot — a new login silently kicked the previous device) with a multi-session model backed by a new `ork_session` table: **up to 3 concurrent sessions per user, with least-recently-active (LRU) eviction** when a 4th device logs in.

All login entry points create sessions; the per-request check and the JSON-API authorizer validate tokens against `ork_session` on every request; logout tears down its own session. `ork_mundane.token` is retained as a dual-written vestigial pointer so the change is revertible in one file.

## What changed

- **New `ork_session` table** (`db-migrations/2026-07-13-add-ork-session-table.sql`) — one row per live token; unique `token`, indexed on `mundane_id`/`expires`.
- **All login paths** (password, IDP/OAuth, self-registration) mint a session via `Authorization::CreateSession()` — cap 3, LRU eviction, and a verified INSERT (retries once, logs, and fails the login rather than returning a token that can't authenticate).
- **Per-request validation** (`class.Controller.php`) and **JSON-API auth** (`IsAuthorized_h`) now validate against `ork_session` on **every** request (the prior PHP-session auth cache no longer short-circuits), so eviction/expiry takes effect immediately — including on AJAX endpoints.
- **Server-side session expiry** is now enforced (72h, slid forward on activity, throttled to ≤1 write/60s; previously effectively cookie-only).
- **Logout** deletes its `ork_session` row and clears the vestigial pointer to prevent replay.

## Deploy

Run the migration once:

```
docker exec -i <db-container> mariadb -u root -p<pw> ork < db-migrations/2026-07-13-add-ork-session-table.sql
```

The migration **backfills currently-active tokens** (unexpired only), so existing logged-in users are **not** kicked on deploy. It is idempotent (`INSERT IGNORE`).

## Testing

No unit-test harness exists in this repo; verified via curl + `mariadb` integration against the local Docker app:

- 4-device login → exactly 3 sessions survive, oldest device kicked (`session_replaced`)
- expired session rejected; valid AJAX still authorizes; logout frees a slot
- a 6-lens review + an adversarial critic pass hardened the session helpers (always-revalidate, verified INSERT with retry/log, CSPRNG token rotation, DRY of the vestigial-token write) and caught/fixed a danger-audit actor-attribution regression

## Coverage

**All** auth entry points are multi-device: password + IDP/OAuth + self-registration (fresh login), the per-request/JSON-API validation, logout, **and** the SOAP/mobile token-refresh path (`Authorize_h` token branch, which now validates against `ork_session` instead of the single `ork_mundane.token`). `ork_mundane.token` remains only as a vestigial rollback pointer.

https://claude.ai/code/session_01DXQ2YePHm59VSzozsxbNXJ


baltinerdist:

Synced up to `master` @ `6ab028a3` (merge commit `74cce921`). This branch was **332 commits behind**.

**Conflicts: 2 files, both security-relevant, resolved deliberately.**

**1. `system/lib/system/class.Controller.php`** — both sides PSR-12'd the whole file *and* both rewrote the same per-request stale-session check. Upstream extracted it into `Model_SessionToken → SessionToken::ValidateSessionToken($uid, $token)` (a lift-and-shift of the old `ork_mundane.token` compare); this branch replaced it with `ork_session` validation.

Resolved by taking upstream's file verbatim — keeping all of upstream's other work in it (What's New wiring, `Model_Player::get_viewer_preferences`/`get_home_kingdom`, `Model_Authorization::has_authority`/`clear_db_after_auth_checks`, `Model_Event::get_rsvp_total_counts_batch`, the new `Controller_EventEmbed` skip-list entry) — and swapping only the check body back to this branch's version.

Taking upstream's version here would have **silently killed the feature**: `ork_mundane.token` holds a single token, so with 3 concurrent device sessions every device but the most recent would get bounced to `msg=session_replaced` on every page load. The owner match (`$_owner === 0 || $_owner !== $_uid_check`) is preserved, so validation is not weakened.

**2. `system/lib/ork3/class.DangerAudit.php`** — both sides PSR-12'd the file; upstream additionally added ~330 lines of new admin audit-log methods (`ListAuditLog`, `ListAuditMethods`, the PR #492 method whitelist). This branch's only semantic change was one line. Took upstream's file verbatim and re-applied the branch's actor-attribution fix:

```php
$mundane_id = (int)($_SESSION['is_authorized_mundane_id'] ?? 0);
// was: Ork3::$Lib->authorization->IsAuthorized($_SESSION['is_authorized_mundane_id']);
```

That line is load-bearing here — it drops the `IsAuthorized_h` session-cache short-circuit so eviction takes effect every request, and `IsAuthorized()` expects a *token*, not a mundane_id. Leaving upstream's line would write `by_whom_id = 0` on every danger-audit row.

**Method:** `git merge upstream/master`, not a rebase — this branch sits behind prior merge commits, and a linear rebase would replay them against the new base and manufacture conflicts that aren't real.

**Verification:** no conflict markers; `php -l` clean on all 242 changed `.php` files across both merge parents; `class.Authorization.php` came through byte-identical to the branch tip (upstream has made no commits to it since the merge base) with **no `true ||` bypass** and both `KeyExists` guards intact; `ValidateSessionByToken` confirmed still using a bound `:token` parameter with `expires > NOW()`. `class.Player.php` auto-merged cleanly with the self-registration auto-login (`CreateSession($new_mundane_id)`) surviving.

Not run: PHPUnit / Playwright — no test DB in the sync worktree.

---

⚠️ **One judgment call worth a maintainer's eye.** Upstream's new `orkui/model/model.SessionToken.php` and `system/lib/ork3/class.SessionToken.php` are now **unreferenced by production code** on this branch — only `tests/Integration/SessionTokenTest.php` still uses them.

`class.SessionToken.php` was deliberately *not* rewritten to delegate to `ork_session`, for two reasons: it would break upstream's two characterization tests (they seed `ork_mundane.token` with no `ork_session` row), and it would mean editing a file that never conflicted. The trade-off is that the Controller now calls `Ork3::$Lib->authorization->…` directly, which cuts slightly against upstream's Megiddo direction of routing Controller reads through models — though upstream's own discovery doc (`ds-13-infrastructure-discovery.md`) proposes exactly an `Authorization::ValidateSessionToken`, so the domain placement matches the stated intent.

If you'd prefer the model layering, the follow-up is: point `SessionToken::ValidateSessionToken` at `ork_session` and update `SessionTokenTest` to seed session rows. Happy to do that as a separate commit if you want it before merge.

PR #492, opened 2026-07-19
Megiddo: frontend domain refactor + development tooling

esdraelon:
## Summary
- **Frontend refactor:** Move business logic and database access out of `orkui/` into domain services / models (Megiddo R-* stack through residual lib work, idiom alignment, Phase 3 gate fixes, and post-refactor rebases onto master).
- **Development tooling:** Test sandbox / ork-db, fuzzy-validator (including v2 classified drift overlays and agent skills), characterization + Playwright coverage, Infection / php-cs-fixer under `tools/`, and Megiddo docs/skills for the process.

## Test plan
- [ ] PHPUnit + relevant e2e for touched domains
- [ ] `bin/ork-db deploy-sandbox` / dual-profile fuzzy validate as needed
- [ ] `cd tools/fuzzy-validator/python && pytest tests/ --cov-fail-under=95`
- [ ] Spot-check no new `$DB` / domain `Ork3::$Lib` in `orkui/` for migrated surfaces

PR #493, opened 2026-07-19
Megiddo: frontend model hop + Player aggregate APIs

esdraelon:
## Summary
Stacked on [#492](https://github.com/amtgard/ORK3/pull/492) (`megiddo/fuzzy-validator-v2`). Merge **#492 first**, then this PR (or retarget base to `master` after #492 lands).

- **Bootstrap / audit model hop:** Route `class.Controller` session/prefs/What’s New/home kingdom and `$DB` clear through models; replace controller `(new Dangeraudit())->audit` with `Authorization->audit`; `index.php` Health/Event via `Model_Health` / `Model_Event`. Extends static gates so Lib / inline Dangeraudit cannot creep back into the frontend hot path.
- **Player aggregates (P3-R0…R4):** First-class domain/model APIs for highest class level, milestones, award/ladder maps & progress, and reconcile smart-rank; thin `Controller_Player` and revised Player templates. Plan under `docs/megiddo/refactor/player-aggregates/`.

## Test plan
- [ ] PHPUnit green (Player / ClassLevel / reconcile coverage as applicable)
- [ ] `rg '\$DB->|Ork3::\$Lib' orkui/` and `rg 'Ork3::\$Lib' system/lib/system/class.Controller.php` clean
- [ ] `rg '\(new Dangeraudit\(\)\)' orkui/controller/` clean
- [ ] Fuzzy `player-profile` test + mirror (or re-check locally)
- [ ] Smoke Player profile + reconcile UI after merge

## Notes
- orkservice JSON registration for aggregates left optional / not required for UI thinness.
- Human Phase 3 close-out (P3-4 / P3-5 / P3-6) still open after this stack.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/megiddo/refactor/idioms-00-charter.md
+++ b/docs/megiddo/refactor/idioms-00-charter.md
@@ -59,7 +59,7 @@ Read these before editing any hop scope. Match **the file being edited** first; 
 
 - Auth checks: `$this->Authorization->has_authority($uid, AUTH_*, $id, $role)` after `load_model('Authorization')`.
 - Auth mutations: `$this->Authorization->add_auth([…])` / `del_auth([…])`.
-- **Dangeraudit:** when `addauth` paths in a file already call `(new Dangeraudit())->audit(…)` after `add_auth`, keep that shape; when file peers route audit through `Model_Authorization`, match peers. Do not change audit payload fields or timing.
+- **Dangeraudit:** `$this->Authorization->audit(…)` after mutations (model wraps `Dangeraudit`). Do **not** `(new Dangeraudit())` in controllers. Do not change audit payload fields or timing.
 
 ### 1.4 JSON shapes and HTTP semantics
 
@@ -76,7 +76,7 @@ Read these before editing any hop scope. Match **the file being edited** first; 
 ### 1.6 Templates and `index.php`
 
 - Template hops (I-15, I-17, I-18): idiom = precomputed auth flags and helper calls consistent with sibling templates; no markup semantic changes.
-- `index.php`: bootstrap calls match R-13/R-14 established patterns (`Health::PingDb`, request-scoped session).
+- `index.php`: bootstrap via models (`Model_Health::ping_db`, `Model_Event::get_event_summary_for_redirect`), not bare `new Health()` / `new Event()`.
 
 ---
 
@@ -89,14 +89,23 @@ Catalog from R-19* execution and agent refactors. Fix only when the **file's dom
 | `(new Model_Player())` in AJAX | `PlayerAjax` username check | `$this->load_model('Player');` → `$this->Player->…` |
 | `(new EventPlanning())` in controller | `EventAjax::remove_rsvp` | `$this->load_model('EventPlanning');` → `$this->EventPlanning->…` |
 | `(new KingdomProfile())` after `load_model` | `KingdomAjax` suspension / abbr | `$this->KingdomProfile->…` (model already loaded) |
-| `(new Dangeraudit())->audit` beside `add_auth` | `KingdomAjax`, `EventAjax`, `AdminAjax`, `ParkAjax` | Keep if all `addauth` branches in file use inline audit; else add `Model_Authorization` wrapper if peers do |
+| `(new Dangeraudit())->audit` in controller | any `*Ajax` / `Unit` | `$this->Authorization->audit(…)` |
+| `Ork3::$Lib->…` in `class.Controller` or `orkui/` | bootstrap prefs / session | `$this->load_model(…)` → snake_case wrapper |
 | `(new Heraldry())->SetEventHeraldry` in controller | `EventAjax` | `$this->load_model('EventPlanning')` or heraldry wrapper on model used elsewhere in file |
 | `(new Weather())` in `Controller_Admin` | weather stats / refresh | `$this->load_model('Weather')` or `Model_AdminDashboard` wrapper |
 | `new SearchService()` only in new model while siblings use `JSONModel` | `Model_Search` (R-11) | Match `model.Event.php` / domain wrapper style in same domain |
 | Mixed `[]` and `array()` within one method | various AJAX | Match **method's** nearest siblings, not PSR-12 |
 | Reformatting entire file for tabs/spaces | any | **Forbidden** — touch only idiom lines |
 
-**Static isolation (non-negotiable):** `rg '\$DB->' orkui/` and `rg 'Ork3::\$Lib' orkui/` must remain zero after every hop.
+**Static isolation (non-negotiable):**
+
+```bash
+rg '\$DB->' orkui/                                    # expect: no matches
+rg 'Ork3::\$Lib' orkui/                              # expect: no matches
+rg 'Ork3::\$Lib' system/lib/system/class.Controller.php   # expect: no matches
+rg '\(new Dangeraudit\(\)\)' orkui/controller/       # expect: no matches
+```
+
 
 ---
 
@@ -145,8 +154,10 @@ Run from repo root. **Exit 0** = pass for PHPUnit; **exit 1 (no matches)** = pas
 ### 4.1 Static isolation (every I-* hop)
 
 ```bash
-rg '\$DB->' orkui/                    # expect: no matches
-rg 'Ork3::\$Lib' orkui/              # expect: no matches
+rg '\$DB->' orkui/                                          # expect: no matches
+rg 'Ork3::\$Lib' orkui/                                    # expect: no matches
+rg 'Ork3::\$Lib' system/lib/system/class.Controller.php     # expect: no matches
+rg '\(new Dangeraudit\(\)\)' orkui/controller/             # expect: no matches
 ```
 
 ### 4.2 Controller idiom drift
@@ -156,7 +167,7 @@ rg 'Ork3::\$Lib' orkui/              # expect: no matches
 rg '\(new Model_' orkui/controller/
 
 # Inline domain construction in controllers (prefer load_model + model wrapper)
-rg '\(new (Player|EventPlanning|KingdomProfile|Heraldry|Weather|Dangeraudit|Report|SearchService)\(\)' orkui/controller/
+rg '\(new (Player|EventPlanning|KingdomProfile|Heraldry|Weather|Report|SearchService)\(\)' orkui/controller/
 
 # Model_* instantiated without load_model in AJAX controllers
 rg 'new Model_' orkui/controller/
--- a/orkui/controller/controller.AdminAjax.php
+++ b/orkui/controller/controller.AdminAjax.php
@@ -60,7 +60,7 @@ class Controller_AdminAjax extends Controller
             $authId = (int)($r['Detail'] ?? 0);
             $this->load_model('Player');
             $persona = $this->Player->get_persona($mid);
-            (new Dangeraudit())->audit('Authorization::AddAuthorization', ['MundaneId' => $mid, 'Type' => AUTH_ADMIN, 'Id' => 0, 'Role' => AUTH_ADMIN], 'Player', $mid, null, [
+            $this->Authorization->audit('Authorization::AddAuthorization', ['MundaneId' => $mid, 'Type' => AUTH_ADMIN, 'Id' => 0, 'Role' => AUTH_ADMIN], 'Player', $mid, null, [
                 'authorization_id' => $authId,
                 'mundane_id'       => $mid,
                 'park_id'          => 0,
--- a/orkui/controller/controller.Player.php
+++ b/orkui/controller/controller.Player.php
@@ -434,7 +434,7 @@ class Controller_Player extends Controller
             'TotalAttendance'   => 0,
             'TotalAwards'       => 0,
             'TotalTitles'       => 0,
-            'HighestClassLevel' => 0,
+            'HighestClassLevel' => $this->Player->get_highest_class_level($id),
             'LastPlayedClass'   => '',
         ];
         if (is_array($this->data['Details']['Awards'])) {
@@ -443,27 +443,6 @@ class Controller_Player extends Controller
                     $this->data['Stats']['TotalAwards']++;
                 } else {
                     $this->data['Stats']['TotalTitles']++;
-                }
-            }
-        }
-        if (is_array($this->data['Details']['Classes'])) {
-            foreach ($this->data['Details']['Classes'] as $c) {
-                $credits = $c['Credits'] + $c['Reconciled'];
-                if ($credits >= 53) {
-                    $lvl = 6;
-                } elseif ($credits >= 34) {
-                    $lvl = 5;
-                } elseif ($credits >= 21) {
-                    $lvl = 4;
-                } elseif ($credits >= 12) {
-                    $lvl = 3;
-                } elseif ($credits >= 5) {
-                    $lvl = 2;
-                } else {
-                    $lvl = 1;
-                }
-                if ($lvl > $this->data['Stats']['HighestClassLevel']) {
-                    $this->data['Stats']['HighestClassLevel'] = $lvl;
                 }
             }
         }
--- a/orkui/controller/controller.Player.php
+++ b/orkui/controller/controller.Player.php
@@ -509,6 +509,7 @@ class Controller_Player extends Controller
             'Awards' => $__awards,
         ]);
         $this->data['ClassParagonMap'] = $this->Player->get_class_paragon_map();
+        $this->data['ClassLevelThresholds'] = $this->Player->get_class_level_thresholds();
         $this->data['LadderMasterMap'] = $this->Player->get_ladder_master_map();
         $this->data['KnightAwardMap'] = $this->Player->get_knight_award_map();
 
--- a/orkui/controller/controller.Player.php
+++ b/orkui/controller/controller.Player.php
@@ -492,160 +492,25 @@ class Controller_Player extends Controller
         $this->data['PlayerTitles'] = $beltline['Titles'];
 
         // ===== Milestones Timeline Data =====
-        $__milestones = [];
         $__awards = is_array($this->data['Details']['Awards']) ? $this->data['Details']['Awards'] : [];
-        $__classes = is_array($this->data['Details']['Classes']) ? $this->data['Details']['Classes'] : [];
-
-        // 1. First Sign-In — use PlayerSinceDate (already computed via MIN(date)
-        // query at controller line ~358); no full-attendance scan needed.
-        $__earliestDate = $this->data['Player']['PlayerSinceDate'] ?? null;
-        if ($__earliestDate && $__earliestDate !== '0000-00-00' && $__earliestDate !== '1970-01-01') {
-            $__milestones[] = ['type' => 'first_signin', 'date' => $__earliestDate, 'icon' => 'fa-door-open', 'description' => 'First sign-in at Amtgard'];
-        }
-
-        // 2. Reached Level 6 in Class — computed client-side once attendance loads
-        // (see PlayerAjax/attendance handler in Playernew_index.tpl). Server-side
-        // generation removed so we don't have to fetch full attendance during
-        // page render.
-
-        // 3-6: Awards-based milestones
-        $__knightIds  = [17, 18, 19, 20, 245];
-        $__knightNames = [17 => 'Flame', 18 => 'Crown', 19 => 'Serpent', 20 => 'Sword', 245 => 'Battle'];
-        $__masterIds  = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 240, 244]; // mirrors $pnOrderToMaster values in Playernew_index.tpl
-        foreach ($__awards as $__aw) {
-            $__aid = (int)($__aw['AwardId'] ?? 0);
-            $__awDate = $__aw['Date'] ?? '';
-            // Prefer the player-specific custom_name when present (Custom Title /
-            // Custom Award rows). Otherwise fall back to the kingdomaward name,
-            // then the underlying award name.
-            $__awName = !empty($__aw['CustomAwardName']) ? $__aw['CustomAwardName'] : (!empty($__aw['KingdomAwardName']) ? $__aw['KingdomAwardName'] : ($__aw['Name'] ?? ''));
-            $__officerRole = $__aw['OfficerRole'] ?? 'none';
-            $__isTitle = (int)($__aw['IsTitle'] ?? 0);
-            $__aliasPeerage = $__aw['AliasPeerage'] ?? '';
-
-            if (empty($__awDate) || $__awDate === '0000-00-00') {
-                continue;
-            }
-
-            // Knight
-            if (in_array($__aid, $__knightIds)) {
-                $__knLabel = isset($__knightNames[$__aid]) ? 'Knight of the ' . $__knightNames[$__aid] : 'Knighted';
-                $__milestones[] = ['type' => 'knight', 'date' => $__awDate, 'icon' => 'fa-shield-alt', 'description' => 'Earned ' . $__knLabel];
-            }
-
-            // Master title — only when the player actually holds the formal Master award,
-            // not merely because they reached rank 10 of the corresponding Order.
-            if (in_array($__aid, $__masterIds)) {
-                $__milestones[] = ['type' => 'master', 'date' => $__awDate, 'icon' => 'fa-star', 'description' => 'Earned ' . $__awName];
-            }
-
-            // Paragon (class-specific paragon awards)
-            $__paragonIds = [37,38,39,40,41,241,42,43,44,45,46,47,242,49,50,51];
-            if (in_array($__aid, $__paragonIds)) {
-                $__milestones[] = ['type' => 'paragon', 'date' => $__awDate, 'icon' => 'fa-gem', 'description' => 'Earned ' . $__awName];
-            }
-
-            // Title (IsTitle=1 and OfficerRole is none, exclude paragons/knights already handled above).
-            // For Custom Titles aliased to a beltline peerage (Page/Squire/etc.),
-            // suppress this — the 'became_associate' milestone already covers it.
-            if ($__isTitle === 1 && in_array($__officerRole, ['none', null]) && !in_array($__aid, $__paragonIds) && !in_array($__aid, $__knightIds)
-                && !in_array($__aliasPeerage, ['Page', 'Lords-Page', 'Squire', 'Man-At-Arms'])) {
-                $__milestones[] = ['type' => 'title', 'date' => $__awDate, 'icon' => 'fa-crown', 'description' => 'Earned the title ' . $__awName];
-            }
-
-            // Served as Officer (OfficerRole is not none)
-            if (!in_array($__officerRole, ['none', null, ''])) {
-                $__milestones[] = ['type' => 'officer', 'date' => $__awDate, 'icon' => 'fa-landmark', 'description' => 'Served as ' . $__awName];
-            }
-        }
-
-        // 7. Became Associate (peerage awards given TO this player - from BeltlinePeers data)
-        $__blPeerLabels = ['Squire' => 'Squire', 'Man-At-Arms' => 'Person-at-Arms', 'Lords-Page' => "Lord's Page", 'Page' => 'Page'];
-        if (!empty($this->data['BeltlinePeers'])) {
-            foreach ($this->data['BeltlinePeers'] as $__bp) {
-                $__peerDate = $__bp['Date'] ?? '';
-                if (empty($__peerDate) || $__peerDate === '0000-00-00') {
-                    continue;
-                }
-                $__peerLabel = $__blPeerLabels[$__bp['Peerage']] ?? $__bp['Peerage'];
-                $__milestones[] = ['type' => 'became_associate', 'date' => $__peerDate, 'icon' => 'fa-handshake', 'description' => 'Became ' . $__peerLabel . ' to ' . $__bp['Persona']];
-            }
-        }
-
-        // 8. Took Associate (peerage awards given BY this player - from BeltlineAssociates data)
-        if (!empty($this->data['BeltlineAssociates'])) {
-            foreach ($this->data['BeltlineAssociates'] as $__ba) {
-                $__assocDate = $__ba['Date'] ?? '';
-                if (empty($__assocDate) || $__assocDate === '0000-00-00') {
-                    continue;
-                }
-                $__assocLabel = $__blPeerLabels[$__ba['Peerage']] ?? $__ba['Peerage'];
-                $__milestones[] = ['type' => 'took_associate', 'date' => $__assocDate, 'icon' => 'fa-hand-holding-heart', 'description' => 'Took ' . $__ba['Persona'] . ' as ' . $__assocLabel];
-            }
-        }
-
-        // 9. Custom milestones from DB
         $__customMs = $this->Player->get_custom_milestones((int)$id);
-        if (is_array($__customMs)) {
-            foreach ($__customMs as $__cm) {
-                $__milestones[] = [
-                    'type' => 'custom',
-                    'date' => $__cm['MilestoneDate'],
-                    'icon' => $__cm['Icon'],
-                    'description' => $__cm['Description'],
-                    'milestoneId' => (int)$__cm['MilestoneId'],
-                ];
-            }
-        }
-
-        // Cross-type dedup:
-        // 1. Remove 'title' milestones for peerage terms (already covered by 'became_associate')
-        // 2. Remove 'title' milestones for "Master X" that duplicate an existing 'master' milestone
-        $__masterMsNames = [];
-        foreach ($__milestones as $__m) {
-            if ($__m['type'] === 'master') {
-                $__masterMsNames[] = strtolower(preg_replace('/^Earned (?:Master )?/', '', $__m['description']));
-            }
-        }
-        $__peerageTerms = ['squire', 'man-at-arms', 'person-at-arms', "lord's page", 'page'];
-        $__milestones = array_values(array_filter($__milestones, function ($m) use ($__masterMsNames, $__peerageTerms) {
-            if ($m['type'] !== 'title') {
-                return true;
-            }
-            $__tn = strtolower(preg_replace('/^Earned the title /', '', $m['description']));
-            if (in_array($__tn, $__peerageTerms)) {
-                return false;
-            }
-            if (substr($__tn, 0, 7) === 'master ') {
-                $__kw = substr($__tn, 7);
-                foreach ($__masterMsNames as $__mn) {
-                    if (strpos($__mn, $__kw) !== false) {
-                        return false;
-                    }
-                }
-            }
-            return true;
-        }));
-
-        // Deduplicate milestones with same description + date
-        $__seen = [];
-        $__milestones = array_filter($__milestones, function ($m) use (&$__seen) {
-            $key = $m['date'] . '|' . $m['description'];
-            if (isset($__seen[$key])) {
-                return false;
-            }
-            $__seen[$key] = true;
-            return true;
-        });
-
-        // Sort chronologically ascending
-        usort($__milestones, function ($a, $b) {
-            return strtotime($a['date']) - strtotime($b['date']);
-        });
-
-        $this->data['Milestones'] = $__milestones;
+        $this->data['Milestones'] = $this->Player->get_player_milestones([
+            'MundaneId' => (int)$id,
+            'PlayerSinceDate' => $this->data['Player']['PlayerSinceDate'] ?? null,
+            'Awards' => $__awards,
+            'BeltlinePeers' => $this->data['BeltlinePeers'] ?? [],
+            'BeltlineAssociates' => $this->data['BeltlineAssociates'] ?? [],
+            'IncludeCustom' => true,
+        ]);
         $this->data['CustomMilestones'] = is_array($__customMs) ? $__customMs : [];
         $this->data['MilestoneConfig'] = $this->data['Player']['MilestoneConfig'] ?? '';
+        $this->data['LadderProgress'] = $this->Player->get_ladder_progress([
+            'MundaneId' => (int)$id,
+            'Awards' => $__awards,
+        ]);
+        $this->data['ClassParagonMap'] = $this->Player->get_class_paragon_map();
+        $this->data['LadderMasterMap'] = $this->Player->get_ladder_master_map();
+        $this->data['KnightAwardMap'] = $this->Player->get_knight_award_map();
 
         // Collapse the Peers/Associates *display* lists to one row per
         // counterparty, keeping the highest-precedence peerage (the SQL
--- a/orkui/controller/controller.Player.php
+++ b/orkui/controller/controller.Player.php
@@ -512,6 +512,11 @@ class Controller_Player extends Controller
         $this->data['LadderMasterMap'] = $this->Player->get_ladder_master_map();
         $this->data['KnightAwardMap'] = $this->Player->get_knight_award_map();
 
+        $__reconcileHints = $this->Player->get_reconcile_suggestions($__awards);
+        $this->data['HasHistoricalLadder'] = !empty($__reconcileHints['HasHistoricalLadder']);
+        $this->data['HasHistorical'] = !empty($this->data['canManageAwards']) && $this->data['HasHistoricalLadder'];
+        $this->data['HasHistoricalTip'] = $this->data['HasHistoricalLadder'];
+
         // Collapse the Peers/Associates *display* lists to one row per
         // counterparty, keeping the highest-precedence peerage (the SQL
         // already orders Squire→Page so the first row wins). Run this AFTER
@@ -621,8 +626,19 @@ class Controller_Player extends Controller
         }
         $this->data['PreloadOfficers'] = $preloadOfficers;
 
-        // AwardId → KingdomAwardId map for current kingdom (pre-match historical award dropdowns)
-        $this->data['AwardIdToKingdomAwardId'] = $this->Player->get_reconcile_award_map((int)$this->session->kingdom_id);
+        $awards = is_array($this->data['Details']['Awards'] ?? null) ? $this->data['Details']['Awards'] : [];
+        $reconcilePage = $this->Player->get_reconcile_page_data([
+            'MundaneId' => $id,
+            'KingdomId' => (int)$this->session->kingdom_id,
+            'Awards' => $awards,
+        ]);
+        $this->data['HistoricalAwards'] = $reconcilePage['HistoricalAwards'];
+        $this->data['RankSuggestions'] = $reconcilePage['RankSuggestions'];
+        $this->data['RealRanksByAwardId'] = $reconcilePage['RealRanksByAwardId'];
+        $this->data['AwardIdToKingdomAwardId'] = $reconcilePage['AwardIdToKingdomAwardId'];
+        $this->data['AwardTypeCount'] = (int)($reconcilePage['Summary']['AwardTypeCount'] ?? 0);
+        $this->data['TotalCount'] = (int)($reconcilePage['Summary']['TotalCount'] ?? 0);
+        $this->data['HasHistoricalLadder'] = !empty($reconcilePage['HasHistoricalLadder']);
     }
 
     private function award_rec_can_delete(int $uid, string $role): bool
--- a/orkui/index.php
+++ b/orkui/index.php
@@ -6,7 +6,8 @@ define('UIR', HTTP_UI_REMOTE . 'index.php?Route=');
 ini_set("error_reporting", E_ALL & ~E_DEPRECATED & ~E_NOTICE & ~E_WARNING);
 
 if (isset($_REQUEST['Route']) && $_REQUEST['Route'] === 'Health') {
-    $ok = (new Health())->PingDb();
+    require_once DIR_MODEL . 'model.Health.php';
+    $ok = (new Model_Health())->ping_db();
     http_response_code($ok ? 200 : 503);
     if ($_SERVER['REQUEST_METHOD'] !== 'HEAD') {
         header('Content-Type: text/plain');
@@ -62,7 +63,8 @@ foreach ($_legacyRedirects as $_old => $_new) {
 // Redirect Event/index/{id} to the kingdom event attendance report with the event name as filter
 if (preg_match('#^Event/index/(\d+)#i', $_REQUEST['Route'], $_m)) {
     $_event_id = (int)$_m[1];
-    $_summary = (new Event())->GetEventSummaryForRedirect($_event_id);
+    require_once DIR_MODEL . 'model.Event.php';
+    $_summary = (new Model_Event())->get_event_summary_for_redirect($_event_id);
     if ($_summary['KingdomId'] > 0 && $_summary['Name'] !== '') {
         header('Location: ' . UIR . 'Reports/event_attendance/Kingdom/' . $_summary['KingdomId'] . '&filter=' . rawurlencode($_summary['Name']), true, 302);
         exit;
--- a/orkui/model/model.Authorization.php
+++ b/orkui/model/model.Authorization.php
@@ -4,7 +4,7 @@ class Model_Authorization extends Model
 {
     public function __construct()
     {
-        parent::__construct($call, $method);
+        parent::__construct();
         $this->Authorization = new APIModel('Authorization');
     }
 
@@ -28,8 +28,30 @@ class Model_Authorization extends Model
         return $this->_authorization_gate()->check($uid, $type, $id, $role);
     }
 
+    /**
+     * First-class audit write for auth / staff mutations (replaces inline new Dangeraudit()).
+     */
+    public function audit($call, $parameters, $entity, $entity_id, $prior_state = null, $post_state = null)
+    {
+        return $this->_dangeraudit()->audit($call, $parameters, $entity, $entity_id, $prior_state, $post_state);
+    }
+
+    /**
+     * HasAuthority / auth ORM shares the global DB connection — clear after nav auth
+     * checks so subclass actions start clean.
+     */
+    public function clear_db_after_auth_checks(): void
+    {
+        $this->_authorization_gate()->clearSharedDb();
+    }
+
     private function _authorization_gate(): AuthorizationGate
     {
         return new AuthorizationGate();
+    }
+
+    private function _dangeraudit(): Dangeraudit
+    {
+        return new Dangeraudit();
     }
 }
--- a/orkui/model/model.Player.php
+++ b/orkui/model/model.Player.php
@@ -322,6 +322,53 @@ class Model_Player extends Model
         return $this->_player()->GetReconcileAwardMap($kingdom_id);
     }
 
+    /**
+     * Pure historical partition + smart-rank (no kingdom map / no DB).
+     *
+     * @param list<array<string, mixed>> $awards
+     * @return array{
+     *   HistoricalAwards: list<array<string, mixed>>,
+     *   RankSuggestions: array<int, int>,
+     *   RealRanksByAwardId: array<int, list<int>>,
+     *   Summary: array{AwardTypeCount: int, TotalCount: int},
+     *   HasHistoricalLadder: bool
+     * }
+     */
+    public function get_reconcile_suggestions(array $awards): array
+    {
+        return $this->_player()->GetReconcileSuggestions($awards);
+    }
+
+    /**
+     * @param array<string, mixed> $request
+     * @return array{
+     *   HistoricalAwards: list<array<string, mixed>>,
+     *   RankSuggestions: array<int, int>,
+     *   RealRanksByAwardId: array<int, list<int>>,
+     *   AwardIdToKingdomAwardId: array<int, int>,
+     *   Summary: array{AwardTypeCount: int, TotalCount: int},
+     *   HasHistoricalLadder: bool
+     * }
+     */
+    public function get_reconcile_page_data(array $request): array
+    {
+        $empty = [
+            'HistoricalAwards' => [],
+            'RankSuggestions' => [],
+            'RealRanksByAwardId' => [],
+            'AwardIdToKingdomAwardId' => [],
+            'Summary' => ['AwardTypeCount' => 0, 'TotalCount' => 0],
+            'HasHistoricalLadder' => false,
+        ];
+        $response = $this->_player()->GetReconcilePageData($request);
+        if (($response['Status'] ?? 1) != 0) {
+            return $empty;
+        }
+        $detail = $response['Detail'] ?? null;
+
+        return is_array($detail) ? array_merge($empty, $detail) : $empty;
+    }
+
     public function get_highest_class_level(int $mundane_id): int
     {
         return $this->_player()->GetHighestClassLevel($mundane_id);
--- a/orkui/model/model.Player.php
+++ b/orkui/model/model.Player.php
@@ -322,6 +322,11 @@ class Model_Player extends Model
         return $this->_player()->GetReconcileAwardMap($kingdom_id);
     }
 
+    public function get_highest_class_level(int $mundane_id): int
+    {
+        return $this->_player()->GetHighestClassLevel($mundane_id);
+    }
+
     public function dismiss_whats_new($mundane_id, $version)
     {
         return $this->_player()->DismissWhatsNew((int)$mundane_id, (string)$version);
--- a/orkui/model/model.Player.php
+++ b/orkui/model/model.Player.php
@@ -327,6 +327,23 @@ class Model_Player extends Model
         return $this->_player()->DismissWhatsNew((int)$mundane_id, (string)$version);
     }
 
+    /** @return array{BasicFonts: int, DyslexiaFonts: int} */
+    public function get_viewer_preferences(int $mundane_id): array
+    {
+        return $this->_player()->GetViewerPreferences($mundane_id);
+    }
+
+    public function get_whats_new_seen(int $mundane_id, string $version): bool
+    {
+        return $this->_player()->GetWhatsNewSeen($mundane_id, $version);
+    }
+
+    /** @return array{KingdomId: int, ParentKingdomId: int} */
+    public function get_home_kingdom(int $mundane_id): array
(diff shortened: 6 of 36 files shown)


baltinerdist (commented):
## Adversarial review — ship with fixes

Reviewed this stacked on top of #492 **including the FIX-PR492 commits** (a hostile multi-agent panel: 5 attack premises across the bootstrap hop, the Player domain, and the templates, plus an Amtgard-domain lens, every finding then attacked by an independent refuter; most findings died there).

**The Player-aggregate extraction holds up almost everywhere.** Verified line-by-line against `git show <pre-refactor>:orkui/template/revised-frontend/Playernew_index.tpl` and the deleted `controller.Player.php` blocks:

- `GetLadderProgress` is a faithful move — rank-only dedupe key, `RankSet + UnrankedCount` effective count, `Approx = (effectiveCount > Rank) && !HasMaster`, `Rank = min(maxRank, max(Rank, effectiveCount))`, the Walker-of-the-Middle-Path (31) skip, synthetic master-only tiles, `strcmp(Name)` ordering. `GetLadderMasterMap`'s `MaxRank` values reproduce the old `($aid === 30) ? 12 : 10` literal for every mapped id.
- `GetReconcileSuggestions` / `GetReconcilePageData` reproduce the old algorithm exactly — historical partition, `IsLadder === 1` filter, `AwardId ASC / date ASC` with missing-date-last via `PHP_INT_MAX`, prefer-existing-rank-if-free-else-smallest-free allocation against `realRanks ∪ usedRanks`. The "suggests a rank a real award already occupies" failure mode is still prevented.
- The catalogue maps in `class.Award.php` equal their former template literals exactly (`GetClassParagonMap` = `$pnClassToParagon` for all 16 pairs, `GetKnightAwardMap` keys = `{17,18,19,20,245}`, `GetLadderMasterMap` = `$pnOrderToMaster` + `$pnOrderNames`). No Paragon can attach to the wrong class.
- `ClassLevel::THRESHOLDS = [5,12,21,34,53]` reproduces the old inline ladder at every boundary, and `revised.js`'s descending loop yields the same 1–6 as the old if/else chain. No credit lands on the wrong side of a threshold.
- Bootstrap hop: `$this->Authorization` **is** loaded at `class.Controller.php:98`, before `clear_db_after_auth_checks()` at :119, on every path including skip-token controllers. `Model_Authorization::audit` forwards to `Dangeraudit::audit` with an identical six-parameter signature at all five converted call sites. `include_once('../startup.php')` on `index.php:3` defines `Model`/`APIModel`/`Health` before the Health branch, so the probe still returns 200/503. `parent::__construct($call, $method)` → `parent::__construct()` removes a pre-existing latent bug.
- Every new controller-supplied template var is `isset`/`is_array`-guarded, no dangling references to removed template locals remain, and the new `PnConfig.classLevelThresholds` reads sit inside the existing `typeof PnConfig === 'undefined'` guard.

One behavioral escape survived.

---

### 🟡 Warlord (award 12) entered the milestone master set, and now emits two milestones on the same date

`system/lib/ork3/class.Award.php:74` (`GetMasterAwardIds` flattens `GetLadderMasterMap`, whose entry `27 => ['MasterAwardIds' => [12]]` injects 12) → `system/lib/ork3/class.Player.php:1277`, `:1307`, `:1325`, `:1409-1425`.

Before this PR the milestone master list was the literal `$__masterIds = [1,2,3,4,5,6,7,8,9,10,11,240,244];` in `controller.Player.php` — **13 entries, no 12** — despite a comment claiming it mirrored `$pnOrderToMaster`. Swapping it for `Award::GetMasterAwardIds()` changes the set.

**Trigger:** any player holding award 12 — Warlord, seeded by `Award::create_award('Warlord', 0, 1, 10, 'Master')`, i.e. `is_ladder=0, is_title=1, officer_role='none'`.

**(a) New milestone that never shipped.** Every Warlord holder gains a gold-star "Earned Warlord" card on the About-tab timeline. That may well be desirable, but it silently changes an unbounded set of existing players' displayed history on deploy, inside a PR whose stated contract is a behavior-preserving move.

**(b) Duplicate on the same date.** The same award *also* hits the title branch at `:1325`, because `IsTitle === 1`, `OfficerRole === 'none'`, and 12 is neither a paragon nor a knight id — that branch excludes `$paragonIds` and `$knightIds` but **not** `$masterIds`:

```php
if ($isTitle === 1 && in_array($officerRole, ['none', null], true)
    && !in_array($aid, $paragonIds, true)
    && !in_array($aid, $knightIds, true)
    && !in_array($aliasPeerage, ['Page', 'Lords-Page', 'Squire', 'Man-At-Arms'], true)
) {
```

The cross-type dedupe at `:1409-1425` doesn't catch it either: it only suppresses titles whose lowercased name starts with `'master '`, and `"Earned the title Warlord"` normalizes to `"warlord"`. The exact date+description dedupe misses too, because the strings differ (`"Earned Warlord"` vs `"Earned the title Warlord"`).

So a Warlord sees two adjacent rows on the same date, and the design-modal milestone-type toggles can't hide it — switch off "title" and it survives as "master", switch off "master" and it survives as "title". That's the double-count the dedupe block exists to prevent, landing on the game's most prestigious martial peerage.

**Why the test suite didn't catch it:** `tests/Unit/PlayerMilestonesTest.php::testWarlordMasterAwardIsMilestone` passes `'IsTitle' => 0` for award 12. The seed data cannot produce that state (`create_award('Warlord', 0, 1, 10, 'Master')` sets `is_title=1`), so the fixture asserts the one case that never occurs in real data — and the case that would have failed is the one that does.

**Suggested fix — two parts:**

1. Add `&& !in_array($aid, $masterIds, true)` to the title-branch condition at `class.Player.php:1325`, alongside the existing paragon/knight exclusions. Name-independent, and it also closes the identical **pre-existing** "Battlemaster" (244) duplicate that the `'master '`-prefix dedupe has always missed.
2. Decide explicitly whether Warlord should trigger a master milestone. If yes, ship it as a documented product change with a release-note callout rather than as a side effect of a refactor. If the old omission was intentional, split the sets — e.g. a distinct `GetMilestoneMasterAwardIds()` excluding 12 — so `GetPlayerMilestones` stops inheriting edits made to the shared ladder/UI badge catalogue.

Either way, change the fixture to `IsTitle => 1` so the test exercises real data.

---

### ⚠️ Merge-order hazard

This branch's merge-base with `megiddo/fuzzy-validator-v2` is **before** the 20 `FIX-PR492` commits from the #492 review landed. Merging #493 as-is after #492 will not revert them (git handles that), but any rebase or squash-merge of this stack that resolves against the older base will. Worth merging #492 into this branch explicitly before landing, so the combined state is what gets tested. I did that locally to review it; the only conflict was `orkui/model/model.Event.php`, where both sides append private accessors — keep all three (`_event_domain`, `_planning`, `_embed`).

### Coverage caveats

Nothing here was exercised against a live profile page — the milestone/ladder/reconcile conclusions rest on source diffing plus the new PHPUnit suite, whose Warlord fixture is already known to encode an impossible data state.

system/lib/ork3/class.Player.php
```diff
@@ -1037,6 +1179,413 @@ public function ComputeClassProgress($request)
...
+
+            if (in_array($aid, $knightIds, true)) {
+                $knLabel = isset($knightMap[$aid]) ? 'Knight of the ' . $knightMap[$aid] : 'Knighted';
+                $milestones[] = [
+                    'type' => 'knight',
+                    'date' => $awDate,
+                    'icon' => 'fa-shield-alt',
+                    'description' => 'Earned ' . $knLabel,
+                ];
+            }
```
baltinerdist, line 1305: **1. Warlord (award 12) now emits two contradictory milestones on every holder's profile**

🟠 **high · problem** · _effort: trivial_

**What:** The controller's hand-maintained master list was `[1,2,3,4,5,6,7,8,9,10,11,240,244]` — deliberately without award 12. `Award::GetMasterAwardIds()` flattens `GetLadderMasterMap()` and therefore includes 12 (Warlord). Award 12 is seeded `is_title = 1` (`create_award('Warlord', 0, 1, 10, 'Master')`), and the title-emission branch excludes `$paragonIds` and `$knightIds` but NOT `$masterIds`, so one award row emits both a `master` milestone ("Earned Warlord", fa-star) and a `title` milestone ("Earned the title Warlord", fa-crown) on the same date. The cross-type dedup only suppresses titles whose name literally begins with "master " — "Warlord" does not match, and the final date|description dedup misses because the descriptions differ. The new test `testWarlordMasterAwardIsMilestone` pins `IsTitle => 0`, a fixture the seed data cannot produce, so it asserts `assertCount(1, ...)` on the one case that cannot duplicate.

**Do:** Add `&& !in_array($aid, $masterIds, true)` to the title-emission condition — better still, track emitted AwardIds and suppress a `title` milestone for any award already emitted as knight/master/paragon, replacing the name-prefix regex with an id-based rule. Keep the name-based filter only for Custom Title rows aliased to a masterhood name (different AwardId, as in `testDedupesMasterTitleAgainstMasterAward`). Then change the Warlord fixture to `'IsTitle' => 1` so it reflects real data, and add a matching case for Battlemaster (244), which has the same defect pre-existing.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/ork3/class.Player.php
```diff
@@ -1037,6 +1179,413 @@ public function ComputeClassProgress($request)
...
+            $lp['MaxRank'] = $maxRank;
+            unset($lp['RankSet'], $lp['UnrankedCount']);
+        }
+        unset($lp);
+
+        foreach ($ladderMap as $orderId => $info) {
+            if (isset($progress[$orderId])) {
+                continue;
+            }
+            $hasMaster = false;
```
baltinerdist, line 1547: **2. Synthetic master-only ladder tile prints the stock order name while ranked rows print the kingdom's name**

🟡 **medium · problem** · _effort: medium_

**What:** For ladders the player holds rows in, `GetLadderProgress` resolves the display name `CustomAwardName → KingdomAwardName → Name`, honouring per-kingdom renames. The backfill branch that creates a tile for a masterhood with no ladder rows uses the hardcoded `$info['LadderName']` from `Award::GetLadderMasterMap()`, so a kingdom that renamed an order sees its own name on progressed ladders and the stock Amtgard name on the master-only tile, in the same grid. This behavior moved verbatim from the template, but this PR promotes `GetLadderMasterMap`/`GetClassParagonMap`/`GetKnightAwardMap` to canonical ("Sole order→master source of truth… UI must not fork this map"), ships `classToParagon` to the browser, and does so with no `kingdom_id` parameter and no read of `ork_kingdomaward` — even though the codebase already reads `ka.is_title` per-kingdom authoritatively. `MaxRank` has the same gap: a kingdom ladder outside the map silently falls back to 10.

**Do:** Pass the player's kingdom_id into `GetLadderProgress` and resolve the synthetic tile's name and rank ceiling through the same `ork_kingdomaward` lookup the ranked rows use, falling back to `LadderName`/`MaxRank` only when there is no local override. Longer term, give the three static catalogues an optional `int $kingdomId` overlay so a kingdom that renames, disables, or extends a ladder does not get the stock map reported back at it.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/ork3/class.Player.php
```diff
@@ -243,6 +243,148 @@ public function GetReconcileAwardMap($kingdomId)
...
+                        $realRanksByAwardId[$aid] = [];
+                    }
+                    if ($rank > 0) {
+                        $realRanksByAwardId[$aid][] = $rank;
+                    }
+                }
+            }
+        }
+
+        // Keep only ladder awards — non-ladder (Custom Award etc.) are not reconcilable
```
baltinerdist, line 292: **3. Unreachable non-ladder branch left inside GetReconcileSuggestions' rank loop**

⚪ **low · problem** · _effort: trivial_

**What:** `GetReconcileSuggestions` filters `$historicalAwards` down to rows with `(int)($a['IsLadder'] ?? 0) === 1`, then iterates that same array and checks `if (!$isLadder) { $rankSuggestions[$awardsId] = 0; continue; }`. Every row reaching that loop is guaranteed `IsLadder === 1` by the preceding filter, so the branch can never execute — inert residue from the pre-extraction controller logic.

**Do:** Delete the dead `if (!$isLadder) {...}` branch, or replace it with a comment documenting the invariant, so the next reader isn't misled into thinking non-ladder rows can reach the rank allocator.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/controller/controller.Player.php
```diff
@@ -434,7 +434,7 @@ public function profile($id = null)
             'TotalAttendance'   => 0,
             'TotalAwards'       => 0,
             'TotalTitles'       => 0,
-            'HighestClassLevel' => 0,
+            'HighestClassLevel' => $this->Player->get_highest_class_level($id),
```
baltinerdist, line 437: **4. Hero stat moved to a different cache than the Class Levels tab, adds a query, and changes the failure value — while the PR's own fix ships dead**

🟠 **high · problem** · _effort: small_

**What:** The profile previously computed `Stats['HighestClassLevel']` by looping `$this->data['Details']['Classes']` — the exact rows the Class Levels tab renders. It now calls `get_highest_class_level($id)` → `ComputeClassProgress` → `GetPlayerClasses` (300s ghettocache, busted by reconciliation), while the tab still renders from `GetPlayerProfileDetails` (separate 60s cache, NOT busted by `SetPlayerReconciledCredits`). Right after an officer reconciles credits the hero can read Level 6 while the tab below reads Level 5 for up to 60s. It also costs a redundant Player construction plus a GetPlayerClasses round trip (annotated in-file as ~185ms cold) for data already in memory, and changes the error value: the old loop returned at least 1 whenever any class row existed, `GetHighestClassLevel` returns 0 if `ComputeClassProgress` errors. The PR already contains the intended fix — `Player::HighestClassLevelFromClasses(array $classes)`, the contract's "Option B — byte-identical to today's profile loop" — with four unit tests, no `Model_Player` wrapper, and zero production callers.

**Do:** Add `Model_Player::highest_class_level_from_classes(array $classes)` as the contract specifies and call it from controller.Player.php:437 with the already-loaded `$this->data['Details']['Classes']`. That restores single-source-of-truth with the tab, drops the query, removes the 0-vs-1 divergence, and gives the pure function a caller. Keep `GetHighestClassLevel(int)` for callers that only have a mundane_id (service layer), and record in 02-api-contract.md §B1 which option shipped. If you keep Option A instead, have `SetPlayerReconciledCredits`/`update_class_reconciliation` also call `bustPlayerProfileCaches()` so both caches invalidate together.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/model/model.Player.php
```diff
@@ -322,11 +322,141 @@ public function get_reconcile_award_map($kingdom_id)
...
+     * @param list<array<string, mixed>> $awards
+     * @return array{
+     *   HistoricalAwards: list<array<string, mixed>>,
+     *   RankSuggestions: array<int, int>,
+     *   RealRanksByAwardId: array<int, list<int>>,
+     *   Summary: array{AwardTypeCount: int, TotalCount: int},
+     *   HasHistoricalLadder: bool
+     * }
+     */
+    public function get_reconcile_suggestions(array $awards): array
```
baltinerdist, line 337: **5. Model wrappers swallow the domain Status envelope; the reconcile page renders a failure as "nothing to reconcile"**

🟠 **high · problem** · _effort: small_

**What:** `get_reconcile_page_data`, `get_player_milestones` and `get_ladder_progress` all check `($response['Status'] ?? 1) != 0` and return an empty structure with no logging and no propagation. The controller assigns the empties straight into `$this->data`, and Playernew_reconcile.tpl renders that as the affirmative "No historical awards found — nothing to reconcile" plus "0 awards across 0 types". On a surface whose entire purpose is deciding what rank a player is entitled to, a silent empty is materially worse than an error page — the officer concludes nothing is owed and moves on. The same shape makes the profile's milestone timeline and ladder tiles simply vanish. This failure mode did not exist before the hop, when the template operated directly on `$Details['Awards']` already in hand.

**Do:** Have the wrappers surface failure distinctly — add a `Loaded`/`Error` key to the DTO rather than silently returning `$empty` — set it into `$this->data` in `Controller_Player::reconcile`/`profile`, and have Playernew_reconcile.tpl render an explicit error state ("Could not load this player's award history — do not treat this as an empty record") instead of the clean empty state. Do the same for the milestone and ladder wrappers so a failed aggregate is visibly different from an empty one.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/template/revised-frontend/Playernew_index.tpl
```diff
@@ -28,7 +28,9 @@
 	}
 
 
-	$knightAwardIds = array(17, 18, 19, 20, 245);
+	// Knight AwardIds from domain (Award::GetKnightAwardMap via controller).
+	// Belt IMAGE URLs stay presentation-local (host/path); AwardIds come from domain.
+	$knightAwardIds = array_map('intval', array_keys(is_array($KnightAwardMap ?? null) ? $KnightAwardMap : []));
```
baltinerdist, line 33: **6. Template `?? []` fallbacks turn a missing controller assignment into "this player earned nothing"**

🟡 **medium · problem** · _effort: small_

**What:** Four domain catalogues that were previously compile-time-correct literals are now controller-supplied variables with silent empty fallbacks: `$knightAwardIds = array_keys($KnightAwardMap ?? [])`, `$ClassParagonMap = []`, `$ClassLevelThresholds = []`, `$ladderMasterMap = $LadderMasterMap ?? []`. Each degrades to a wrong-but-plausible render: no knight belts, no paragon badges, and — worst — with empty `classLevelThresholds` the JS computes `thresholds = [0]`, so `lvl` stays 1 for every class, `pct = Math.round(total / (thresholds[1] || 1) * 100)` yields absurd values (e.g. 4700%), `isMax` never fires, and a Level 6 Wizard displays as Level 1. An empty `ladderMasterMap` makes `map[baseAwardId]` undefined so `pnApplyLadderSuggestion` returns early and the "already holds the Master peerage / already at max rank" guard on the recommendation modal silently stops firing.

**Do:** Do not let missing domain data degrade into 'earned nothing'. Assign all four maps from one controller helper, and in the JS bail out of the class-progress card and the level column when `PnConfig.classLevelThresholds` is empty rather than rendering Level 1 for everyone. Gate the ladder-suggestion JS on a `PnConfig.ladderMasterMapLoaded` flag so a missing map disables submission instead of allowing a duplicate Master recommendation, and render a visible 'data unavailable' note so the failure is diagnosable instead of looking like a records problem.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/system/class.Controller.php
```diff
@@ -51,7 +51,8 @@ public function __construct($method = null, $action = null)
         if (!$_skipTokenCheck && isset($this->session->user_id) && isset($this->session->token)) {
```
baltinerdist, line 51: **7. The hop rebuilds the whole Player domain object graph 3-5x on every request, including per-keystroke autocomplete endpoints**

🟠 **high · problem** · _effort: small_

**What:** The old bootstrap read viewer prefs / What's-New / home kingdom off `Ork3::$Lib->player`, a singleton startup.php builds once. Now `Controller::__construct` does `load_model('Player')` (→ `Model::__construct` → `new APIModel('Player')` → `new Player()`), then `get_viewer_preferences()` and `get_whats_new_seen()` each call `Model_Player::_player()`, which is `return new Player();` with no memoization; `Controller::index()` repeats the pattern for `get_home_kingdom`. That is 3 Player constructions per authenticated page and 5 on the home route, where it was previously zero. Each `new Player()` builds 5 yapo objects then `load_model('Kingdom'|'Park'|'Pronoun')`, instantiating Model_Kingdom (7 APIModel/JSONModel wrappers, each constructing its own domain object + yapos), Model_Park (6 more) and Model_Pronoun — on the order of 20+ constructions per call. This block is not gated by the existing `$_skipTokenCheck` allow-list, so EventAjax/KingdomAjax/ParkAjax/AdminAjax pay the full cost on every `playersearch` keystroke to populate template data that `echo json_encode(...); exit;` never reads.

**Do:** Memoize the domain factories in the model layer: `private ?Player $_playerDomain = null; private function _player(): Player { return $this->_playerDomain ??= new Player(); }` (same for `_event_domain`, `_session_token`, `_health`, `_authorization_gate`, `_dangeraudit`) — every call site already `clear()`s its own bound yapo, so sharing one instance per request is safe. Better still, reuse the `$this->Player` APIModel that `Model::__construct` already built rather than constructing a second domain instance, and hoist the three bootstrap reads into a single `Model_Player::get_bootstrap_context($uid)`. Separately, reuse `$_skipTokenCheck` (renamed to something like `$_isJsonController`) to also gate the viewer-preference / What's-New / nav-menu block so JSON-only controllers skip it entirely.

<sub>Anchored to the nearest line inside this PR's diff; the finding refers to `system/lib/system/class.Controller.php:39-119, :145-153; orkui/model/model.Player.php:536-539 (_player factory, 39 call sites); system/lib/ork3/class.Player.php:5-16, :2460 (load_model cascade)`.</sub>

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/controller/controller.Player.php
```diff
@@ -513,160 +492,31 @@ public function profile($id = null)
...
-                if (empty($__assocDate) || $__assocDate === '0000-00-00') {
-                    continue;
-                }
-                $__assocLabel = $__blPeerLabels[$__ba['Peerage']] ?? $__ba['Peerage'];
-                $__milestones[] = ['type' => 'took_associate', 'date' => $__assocDate, 'icon' => 'fa-hand-holding-heart', 'description' => 'Took ' . $__ba['Persona'] . ' as ' . $__assocLabel];
-            }
-        }
-
-        // 9. Custom milestones from DB
         $__customMs = $this->Player->get_custom_milestones((int)$id);
```
baltinerdist, line 496: **8. Custom milestones are now fetched twice on every profile render**

🟡 **medium · problem** · _effort: small_

**What:** `Controller_Player::profile` calls `get_custom_milestones((int)$id)` to populate `$this->data['CustomMilestones']`, then calls `get_player_milestones([... 'IncludeCustom' => true])`, whose body calls `$this->GetCustomMilestones($mundaneId)` again internally. `GetCustomMilestones` is an uncached raw yapo `find()` with no memoization anywhere in the chain, so `ork_player_milestones` is hit twice back-to-back on the highest-traffic page type in the app where it was hit once before. This contradicts the method's own docblock ("Prefer preloaded Awards / beltline to avoid N+1") — Awards/BeltlinePeers/BeltlineAssociates got the preload treatment, the one collection the controller had already loaded did not.

**Do:** Add an optional `'CustomMilestones' => ?list` key to the `GetPlayerMilestones` request shape mirroring `BeltlinePeers`/`BeltlineAssociates`: use it directly when present, fall back to the internal fetch only when absent. Pass `'CustomMilestones' => $__customMs` from the controller, keep `IncludeCustom` as the load-internally fallback, and update 02-api-contract.md §B3 to match.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/index.php
```diff
@@ -6,7 +6,8 @@
 ini_set("error_reporting", E_ALL & ~E_DEPRECATED & ~E_NOTICE & ~E_WARNING);
 
 if (isset($_REQUEST['Route']) && $_REQUEST['Route'] === 'Health') {
```
baltinerdist, line 8: **9. Health and legacy-redirect probes now boot the model layer, one of them before $Settings/$Session exist**

⚪ **low · problem** · _effort: trivial_

**What:** `orkui/index.php` constructs `new Model_Health()` at line 10, but `$Settings`/`$Session` are not created until lines 40-42 — `Model::__construct` does `global $Settings, $Session;` and assigns both null, then builds an `APIModel('Health')` that `ping_db()` never touches because `_health()` constructs a second `Health`. A liveness probe whose whole point is to be minimal now depends on the model constructor contract and builds the domain twice. The redirect path is heavier: `new Model_Event()` for one 302 builds APIModels for Kingdom, Report, Event and Heraldry plus a JSONModel Search (and runs `$this->Log = $LOG;` with no `global $LOG`, silently assigning null) on top of the base `APIModel('Event')`, then `_event_domain()` builds yet another `Event` — roughly seven domain graphs to look up one event name.

**Do:** Use the instance `Model::__construct` already created (`return $this->Health->PingDb();` / `return $this->Event->GetEventSummaryForRedirect($event_id);`) instead of hand-rolled `new Health()`/`new Event()` factories, or give Model_Health its own no-op constructor. For the redirect, put `get_event_summary_for_redirect` on a lean model (or make Model_Event's APIModel members lazy) so a 302 does not pay the full constructor. If the Health probe must stay above the bootstrap, keep it on the domain class — index.php is not a controller and is not what the `orkui/` gate protects.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/system/class.Controller.php
baltinerdist, line 51: **10. The new Model_* indirection — the entire point of the hop — has zero tests**

🟠 **high · problem** · _effort: small_

**What:** All seven characterization tests shipped alongside this work instantiate the pre-existing domain classes directly (`new Health()`, `new SessionToken()`, `new Event()`, `new Player()`) and never call `Model_Health::ping_db()`, `Model_SessionToken::validate_session_token()`, `Model_Event::get_event_summary_for_redirect()`, `Model_Player::get_viewer_preferences()/get_whats_new_seen()/get_home_kingdom()`, or `Model_Authorization::audit()`/`clear_db_after_auth_checks()`. A wiring bug in one of these one-line wrappers (wrong method name, swapped cast, wrong array key) would not fail a single test, even though it is exactly the code this refactor added and exactly the seam it exists to make testable. A repo-wide search of tests/ finds zero references to `Model_Authorization::has_authority` or `clear_db_after_auth_checks`; `AuthorizationLibTest` only exercises `Ork3::$Lib->authorization->HasAuthority(...)` and would stay green through a parameter-order bug in the model path.

**Do:** Add a small PHPUnit suite that instantiates the Model_* classes themselves (reusing the existing InfrastructureFixture/AdminDashboardFixture helpers) and asserts they forward with the correct types and shape: `(new Model_Health())->ping_db()`, `(new Model_SessionToken())->validate_session_token($mid,$token)`, `(new Model_Event())->get_event_summary_for_redirect($id)`, `(new Model_Player())->get_home_kingdom($mid)`/`get_viewer_preferences($mid)`, and `Model_Authorization::audit()`/`has_authority()`. Each is a one-line delegate, so this is cheap and closes the exact gap.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/ork3/class.Player.php
```diff
@@ -243,6 +243,148 @@ public function GetReconcileAwardMap($kingdomId)
...
+     * Reconcile page DTO: suggestions + kingdom award map.
+     *
+     * @param array{
+     *   MundaneId?: int,
+     *   KingdomId?: int,
+     *   Awards?: ?list
+     * } $request
+     * @return array{Status: int, Error?: string, Detail: array<string, mixed>}
+     */
+    public function GetReconcilePageData(array $request)
```
baltinerdist, line 369: **11. GetReconcilePageData — the method actually wired to the reconcile page — is untested**

🟠 **high · problem** · _effort: small_

**What:** `Controller_Player::reconcile` calls `Model_Player::get_reconcile_page_data()`, which forwards to `Player::GetReconcilePageData()`. No test exercises that method; only the inner pure `GetReconcileSuggestions()` is unit tested. The `Success()`/`Detail` wrapping, the merge of `AwardIdToKingdomAwardId` via `GetReconcileAwardMap`, and the DB-fallback branch (fetching Awards via `AwardsForPlayer` when not supplied) are all unverified glue that could silently break the reconcile page while every existing unit test stays green.

**Do:** Add a direct test for `GetReconcilePageData` covering (a) Awards passed explicitly, asserting the merge with `AwardIdToKingdomAwardId` is correct, and (b) the `InvalidParameter` path when MundaneId is invalid and Awards is omitted — which is also the path that currently reaches the silently-empty model wrapper in item 5.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tests/Unit/PlayerHighestClassLevelTest.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public function testHighestClassLevelFromClassesEmptyIsZero(): void
+    {
+        $this->assertSame(0, Player::HighestClassLevelFromClasses([]));
+    }
+
+    public function testHighestClassLevelFromClassesLevelBoundaries(): void
+    {
+        $cases = [
```
baltinerdist, line 31: **12. Class-level boundary tests silently skip every fractional "just below threshold" case they claim to cover**

🟠 **high · problem** · _effort: trivial_

**What:** `PlayerHighestClassLevelTest::testHighestClassLevelFromClassesLevelBoundaries` builds `$cases = [0.0 => 1, 4.9 => 1, 5.0 => 2, 11.9 => 2, ...]` and iterates with `foreach ($cases as $credits => $expectedLevel)`. PHP casts array keys to int, so `4.9` becomes key `4`, `11.9` becomes `11`, `20.9` becomes `20`, `33.9` becomes `33`, `52.9` becomes `52`. The test never exercises the values just below a threshold that its naming implies — it tests values that were never in danger of tripping the `>=` comparison, and would pass even if a future edit shifted a threshold by up to ~1 credit. `ClassLevelTest::testLevelBoundaries` has the identical defect (pre-existing, same one-line fix).

**Do:** Rewrite the fixtures as a list of pairs — `[[0.0, 1], [4.9, 1], [5.0, 2], [11.9, 2], ...]` — and iterate with `foreach ($cases as [$credits, $expectedLevel])` so the fractional values survive and are actually asserted. Apply the same fix to ClassLevelTest while you are in there.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tests/Unit/AwardCatalogueMapsTest.php
```diff
@@ -0,0 +1,68 @@
...
+            18 => 'Crown',
+            19 => 'Serpent',
+            20 => 'Sword',
+            245 => 'Battle',
+        ];
+
+        $this->assertSame($expected, Award::GetKnightAwardMap());
+    }
+
+    public function testMasterAwardIdsFlattenLadderMasterMapIncludingWarlord(): void
```
baltinerdist, line 35: **13. AwardCatalogueMapsTest verifies the derived maps against the same formula the production code uses**

🟡 **medium · problem** · _effort: trivial_

**What:** `testParagonAwardIdsMatchClassParagonValues` computes its expected value with the identical expression `Award::GetParagonAwardIds()` uses internally (`array_values(array_unique(array_map('intval', array_values(Award::GetClassParagonMap()))))`), so the assertion can never fail — it only checks the formula was copy-pasted twice. `testMasterAwardIdsFlattenLadderMasterMapIncludingWarlord` has the same problem in its closing `assertSame`, partially redeemed by the earlier concrete `assertContains` checks. A refactor bug in either derivation (wrong source array, dropped `array_unique`/`array_map`) would sail through.

**Do:** Replace the self-referential expected-value computation with a literal hardcoded array — as the same file already does correctly in `testClassParagonMapMatchesFormerTemplateLiterals` — so the test pins the derived output rather than re-deriving it.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

baltinerdist (commented):
## Distributed Review — PR #493

PR #493's two halves land very differently. The bootstrap model hop is mechanically clean — all four ripgrep gates pass at this SHA and no behavior changed — but it swapped a process-wide `Ork3::$Lib->player` singleton for an unmemoized `new Player()` factory that cascades into three sibling models, so every authenticated page now builds 3-5 full Player object graphs where it previously built zero. The Player aggregate extraction is the higher-risk half: it is mostly faithful, but one catalogue substitution (`$__masterIds = [1..11,240,244]` → `Award::GetMasterAwardIds()`, which adds Warlord/12) produces a duplicate, visually contradictory milestone on every Warlord's public profile, and the new characterization test for that exact award pins `IsTitle => 0`, a shape the seed data (`create_award('Warlord', 0, 1, 10, 'Master')`) cannot produce — so the regression ships green. Separately, `HighestClassLevel` moved off the already-fetched `Details['Classes']` onto a second cache (`Player.GetPlayerClasses`, 300s) while the Class Levels tab still renders from `GetPlayerProfileDetails` (60s), so the hero stat and the tab below it can disagree after a reconcile; the PR already contains the fix (`HighestClassLevelFromClasses`, contract "Option B") but ships it dead with no model wrapper and no caller. Finally, the three new model wrappers convert a non-zero domain Status into an indistinguishable empty result, which the reconcile page renders as the affirmative "No historical awards found — nothing to reconcile." Test coverage is thorough on the pure transforms but has zero tests for the actual new indirection (Model_Health, Model_SessionToken, Model_Player bootstrap methods, Model_Authorization::audit, GetReconcilePageData), and two new boundary tests silently skip their fractional cases because PHP truncates float array keys. Scope note: the unauthenticated AuthorizationGate JSON endpoint and the pre-auth event redirect both come from PR #492 and are excluded.

**28 findings** across 8 themes, from a panel of specialist reviewers (technical + subject-matter). 26 are attached inline below; 2 are cross-cutting and listed here.

Themes: Milestone & ladder aggregate correctness, HighestClassLevel: two sources of truth for one number, Failure handling: empty results are indistinguishable from failures, Bootstrap model-hop performance, Test coverage of the new seam, Reconcile tool explainability and precedence defensibility, Audit hop shape and reliability, Layering, gates, and forward exposure

### Cross-cutting findings (no single line to anchor to)

**18. Reconcile page still forks the ladder MaxRank rule this PR made canonical** — 💡 medium · enhancement · _effort: small_
- **What:** The PR's own contract doc (02-api-contract.md §E, Anti-patterns) states the ladder/order maps must not be forked outside `Award::GetLadderMasterMap()`, and `GetLadderProgress` correctly reads MaxRank from that map. Playernew_reconcile.tpl — a template this PR rewired — still computes `$maxRank = ($aid === 30) ? 12 : 10;` locally to size the rank `<input>` for each historical award row. It agrees with the domain map today (Zodiac=12, everything else=10), but it is a second unmaintained copy of exactly the rule this PR consolidated everywhere else, on the surface where getting MaxRank wrong caps a player's proposed rank.
- **Do:** Have `Controller_Player::reconcile` assign `$this->data['LadderMasterMap'] = $this->Player->get_ladder_master_map();` (already available on the model, already done in `profile`) and have the template read `$ladderMasterMap[$aid]['MaxRank'] ?? 10` instead of the inline ternary.
- **Where:** `orkui/template/revised-frontend/Playernew_reconcile.tpl:185; orkui/controller/controller.Player.php:582-643`

**21. The new `$this->Authorization->audit()` idiom depends on a property one controller redefines to a different type** — 🟡 medium · problem · _effort: small_
- **What:** The six audit call sites rely on `Controller::__construct` having set `$this->Authorization = new Model_Authorization()`. `Controller_Authorization::__construct` reassigns `$this->Authorization = new APIModel('Authorization')` after `parent::__construct()`, so in that controller the property is the *domain* class and `->audit(...)` would fall through `APIModel::__call` to a non-existent `Authorization::audit()` — a fatal. That controller is precisely the one whose `add_auth`/`del_auth` siblings the charter says should carry an audit, so the trap is one commit away. The call sites are also inconsistent about it: AdminAjax, EventAjax::auth, KingdomAjax and ParkAjax call `load_model('Authorization')` immediately before use (redundant — the base constructor already did), while EventAjax::add_staff, EventAjax::remove_staff and Unit rely on the base-constructor instance. Two conventions for one new idiom, one of them silently unsafe.
- **Do:** Make the idiom total: fix `Controller_Authorization` so `$this->Authorization` means the same thing everywhere, pick one of the two shapes (explicit `load_model` at every site, or rely on the base constructor at every site), apply it uniformly, and make the charter's §1.3 idiom text match. Better still, stop overloading a property whose type varies per controller — see item 23.
- **Where:** `orkui/controller/controller.Authorization.php:6-9; call sites at controller.AdminAjax.php:63, controller.EventAjax.php:328/417/479, controller.KingdomAjax.php:596, controller.ParkAjax.php:381, controller.Unit.php:153`

---

Review performed by Claude, take with a grain of salt. -TofH

system/lib/ork3/class.Player.php
```diff
@@ -1037,6 +1179,413 @@ public function ComputeClassProgress($request)
...
+            ) {
+                $milestones[] = [
+                    'type' => 'title',
+                    'date' => $awDate,
+                    'icon' => 'fa-crown',
+                    'description' => 'Earned the title ' . $awName,
+                ];
+            }
+
+            if (!in_array($officerRole, ['none', null, ''], true)) {
```
baltinerdist, line 1338: **14. Officer-role handling is never exercised in either new characterization suite**

🟡 **medium · problem** · _effort: small_

**What:** `GetPlayerMilestones` emits a `type => 'officer'` milestone ("Served as X") whenever OfficerRole is not none/null/'', but every fixture in PlayerMilestonesTest hardcodes `'OfficerRole' => 'none'`, so one of the six milestone types the rewritten method produces is never tested. The same blind spot exists in PlayerReconcileSuggestionsTest: the `row()` helper hardcodes `'OfficerRole' => 'none'`, so the guard that keeps officer records out of ladder reconciliation entirely has no assertion behind it — and misclassifying an officer record as a ladder rank directly affects a player's precedence.

**Do:** Add a milestones case with a non-none OfficerRole (e.g. 'Reeve') asserting a `type => 'officer'` milestone with the expected description/icon, and a reconcile case mixing an officer-role row with real ladder rows asserting it is absent from HistoricalAwards and does not shift RankSuggestions or the Summary counts. While there, add the two realistic data shapes the reconcile suite is missing: two rows sharing the same AwardsId, and two 'real' rows at the same Rank for the same AwardId.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/template/revised-frontend/script/revised.js
```diff
@@ -726,24 +726,24 @@ $(document).ready(function() {
         }
     });
 
-    // ---- Class Level Calculation ----
+    // ---- Class Level Calculation (thresholds from ClassLevel via PnConfig) ----
```
baltinerdist, line 729: **15. revised.js reimplements the class-level algorithm from the thresholds, with no parity test tying it to the PHP**

💡 **medium · enhancement** · _effort: small_

**What:** The PR correctly stops hardcoding the threshold constants in JS and feeds `ClassLevel::THRESHOLDS` through PnConfig, but revised.js still re-derives per-row level and the leveled-up/soon badges with its own descending-scan loop, duplicating the *algorithm* implemented in `ClassLevel::computeClassLevel` and pinned by the PHP boundary table. I verified the two agree today for `[5,12,21,34,53]`, but if the PHP boundary semantics ever change the JS copy diverges silently with nothing in CI to catch it. The same duplicated loop appears twice more inside Playernew_index.tpl.

**Do:** Either have the server compute and hand down each row's Level (removing the client-side reimplementation entirely, which also fixes the empty-thresholds degradation in item 6), or add a lightweight JS test using the same boundary table as ClassLevelTest (0.0=>1, 4.9=>1, 5.0=>2, … 53.0=>6) to guarantee parity.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/ork3/class.Player.php
```diff
@@ -243,6 +243,148 @@ public function GetReconcileAwardMap($kingdomId)
...
+     * @param list<array<string, mixed>> $awards
+     * @return array{
+     *   HistoricalAwards: list<array<string, mixed>>,
+     *   RankSuggestions: array<int, int>,
+     *   RealRanksByAwardId: array<int, list<int>>,
+     *   Summary: array{AwardTypeCount: int, TotalCount: int},
+     *   HasHistoricalLadder: bool
+     * }
+     */
+    public function GetReconcileSuggestions(array $awards): array
```
baltinerdist, line 260: **16. The three genuinely pure aggregates are instance methods on a DB-bound class, so "unit" tests must boot the runtime**

💡 **medium · enhancement** · _effort: medium_

**What:** `GetReconcileSuggestions`, `GetLadderProgress`, and `GetPlayerMilestones` (when Awards/beltline are preloaded and IncludeCustom is false) touch no database at all — they are pure transforms over arrays. But they are instance methods on `Player`, whose constructor builds five yapo objects and loads three sibling models, so every new test does `new Player()` in setUp and tests/bootstrap.php must require startup.php with a live connection. `HighestClassLevelFromClasses` already demonstrates the alternative (static, pure, trivially testable). Given the PR's stated goal of making these testable domain functions, the pure ones are the strongest candidates and the only three still carrying the DB dependency for no reason.

**Do:** Make `GetReconcileSuggestions` and `GetLadderProgress` static (neither uses `$this`), and split `GetPlayerMilestones` into a static `MilestonesFromAggregates(array $awards, ?string $playerSince, array $peers, array $associates, array $customMilestones)` plus a thin instance method that loads the four inputs and delegates. Keep the existing instance signatures as one-line forwarders so no call site changes, then move the three test suites onto the static entry points so they no longer need a database.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/ork3/class.Player.php
```diff
@@ -243,6 +243,148 @@ public function GetReconcileAwardMap($kingdomId)
...
+            if ((int)$a['AwardId'] !== (int)$b['AwardId']) {
+                return (int)$a['AwardId'] - (int)$b['AwardId'];
+            }
+            $da = ($ts = strtotime($a['Date'] ?? '')) > 0 ? $ts : PHP_INT_MAX;
+            $db = ($ts = strtotime($b['Date'] ?? '')) > 0 ? $ts : PHP_INT_MAX;
+
+            return $da - $db;
+        });
+
+        $rankSuggestions = [];
```
baltinerdist, line 308: **17. Rank suggestions expose a bare number while the domain computes — and discards — the full justification**

💡 **high · enhancement** · _effort: small_

**What:** `GetReconcileSuggestions` builds `$groupState[$aid]['realRanks']` (ranks already held by verified records) and `usedRanks` (ranks allocated earlier in the same pass) — exactly the evidence a herald needs to defend a proposed rank — then returns only `RankSuggestions[$awardsId] => int`. The page pre-fills a rank box and the officer has no way to see whether 7 was chosen because the row already carried rank 7 and nothing claimed it, because ranks 1-6 were occupied and 7 was the lowest free slot, or because the row's recorded rank collided with a verified record and was silently overridden. Reconciling a long-serving player's Order of the Warrior history is precisely where a wrong click inflates or strips precedence.

**Do:** Widen the per-row suggestion to `['Rank' => int, 'Reason' => 'kept-existing'|'lowest-free'|'collision-reassigned', 'RecordedRank' => int, 'OccupiedRanks' => list<int>]` — all of which is already in scope in the allocator loop — and render it in Playernew_reconcile.tpl next to the pre-filled rank (e.g. "row records rank 3, but rank 3 is already held by a verified award — proposing 7, the lowest unused"). Keep `RankSuggestions` as-is for back-compat or migrate the template in the same change.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/ork3/class.Player.php
```diff
@@ -1037,6 +1179,413 @@ public function ComputeClassProgress($request)
...
+                }
+                if ($rank > 0) {
+                    $progress[$aid]['RankSet'][$rankKey] = true;
+                } else {
+                    $progress[$aid]['UnrankedCount']++;
+                }
+            }
+        }
+
+        foreach ($progress as $lpAid => &$lp) {
```
baltinerdist, line 1533: **19. Ladder tiles discard the evidence behind the ~ approximation**

💡 **low · enhancement** · _effort: small_

**What:** `GetLadderProgress` computes `RankSet` (distinct recorded ranks) and `UnrankedCount`, uses them to push the displayed rank up to `effectiveCount`, then `unset()`s both. The tile asserts "Rose 5 / 10" with only a `~` and a generic tooltip, so two players show the same number where one has five verified ranked awards and the other has one rank-2 award plus four unranked legacy imports — a distinction that matters for precedence and is exactly the question the reconcile tool exists to answer. The `Approx` flag is also suppressed entirely when the player holds the masterhood, hiding unreconciled rows from the person most likely to fix them.

**Do:** Carry `DistinctRankedCount`, `UnrankedCount` and `HighestRecordedRank` through to the LadderTile DTO (extend 02-api-contract.md §B4) and have the tooltip say '5 shown = 2 recorded ranks + 3 entries with no rank recorded — reconcile to confirm'. Deep-link the tile to Player/reconcile when the viewer can manage awards.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/ork3/class.Player.php
baltinerdist, line 260: **20. Reconcile audit records the rank applied but not the rank the tool proposed**

💡 **low · enhancement** · _effort: small_

**What:** `GetReconcileSuggestions` now computes, as a first-class addressable value, exactly what rank the system believes each historical award should hold, and the page pre-fills the officer's form with it. `Player::ReconcileAward` audits prior/post award state but its `parameters` payload contains only the submitted `Rank` — there is no record of whether the officer accepted the machine suggestion or overrode it. In a precedence dispute ("who decided I was Rose 3 and not Rose 5?") that is the whole question, and the suggestion is not reproducible after the fact because it depends on the award set as it stood at reconcile time.

**Do:** Submit the suggested rank alongside the chosen rank from the reconcile form (e.g. `SuggestedRank`) so it lands in Dangeraudit's `parameters` JSON for `Player::ReconcileAward`, and render it in the Admin audit-log diff as "suggested N, applied M". No schema change needed — `parameters` is already a JSON blob.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/model/model.Authorization.php
```diff
@@ -28,8 +28,30 @@ public function has_authority(int $uid, string $type, $id, ?string $role): bool
         return $this->_authorization_gate()->check($uid, $type, $id, $role);
     }
 
+    /**
```
baltinerdist, line 31: **22. The new first-class audit API returns a value that is always null, takes its actor from ambient $_SESSION, and accepts an unvalidated entity string**

🟡 **medium · problem** · _effort: small_

**What:** `Model_Authorization::audit()` is written as `return $this->_dangeraudit()->audit(...)`, but `Dangeraudit::audit()` has no `return`, so the model method always returns null regardless of whether the row was written — and inside it, `$this->audit->save()` is unchecked and the follow-up entity_id UPDATE runs under ERRMODE_WARNING, so a lost audit row produces no exception, no log, and no signal. The actor is resolved internally from `$_SESSION['is_authorized_mundane_id']`, which the API cannot override — so it can't be called correctly from a test, CLI job, or service entry point, and four of the six migrated controllers are in `$_skipTokenCheck` (they run without stale-session validation) yet perform the highest-privilege grants. The signature is also untyped: `$entity` goes straight to `danger_audit.entity`, while `Dangeraudit::ListAuditLog()` hard-whitelists EntityType to ['Player','Park','Kingdom','Event','Unit'] and silently drops anything else, so a typo writes a row the officer-facing log can never surface.

**Do:** Make `Dangeraudit::audit()` return bool (save() success, and when `$entity_id > 0` the entity_id patch too), have `Model_Authorization::audit(): bool` propagate it, and logtrace a warning on false. Add an explicit `?int $actor_id = null` parameter threaded down to Dangeraudit (falling back to the ambient session lookup only when omitted) and pass `(int)$this->session->user_id` at all six call sites. Type the signature (`audit(string $call, array $parameters, string $entity, int $entity_id, ...)`) and validate `$entity` against a shared `Dangeraudit::ENTITY_TYPES` constant reused by `ListAuditLog()`.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/model/model.Authorization.php
```diff
@@ -28,8 +28,30 @@ public function has_authority(int $uid, string $type, $id, ?string $role): bool
...
+     * First-class audit write for auth / staff mutations (replaces inline new Dangeraudit()).
+     */
+    public function audit($call, $parameters, $entity, $entity_id, $prior_state = null, $post_state = null)
```
baltinerdist, line 34: **23. Audit ownership is split across two unrelated models with duplicate private factories**

💡 **medium · enhancement** · _effort: small_

**What:** `Model_Authorization::audit()` is a pure pass-through to Dangeraudit with no relationship to the Authorization model's own state — it was placed there because the base controller already loads that model, not because authorization owns auditing. The evidence: EventAjax now writes `EventStaff::Add/Update/Remove` audit rows through `Model_Authorization`, while `Model_AdminDashboard` already owns the audit *reads* (`audit_log`, `audit_methods`, `audit_display_maps`) behind its own identical private `_dangeraudit()` factory. One domain, two model owners, two duplicate factories, and a write API named after the wrong bounded context.

**Do:** Introduce a thin `Model_Audit` with one `_dangeraudit()` factory owning both the write (`audit`) and the reads currently on `Model_AdminDashboard`, and point all six controller call sites plus Model_AdminDashboard at it. This also resolves item 21 by removing the overloaded `$this->Authorization` property from the audit path entirely.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/controller/controller.AdminAjax.php
```diff
@@ -60,7 +60,7 @@ public function global($p = null)
             $authId = (int)($r['Detail'] ?? 0);
             $this->load_model('Player');
             $persona = $this->Player->get_persona($mid);
-            (new Dangeraudit())->audit('Authorization::AddAuthorization', ['MundaneId' => $mid, 'Type' => AUTH_ADMIN, 'Id' => 0, 'Role' => AUTH_ADMIN], 'Player', $mid, null, [
+            $this->Authorization->audit('Authorization::AddAuthorization', ['MundaneId' => $mid, 'Type' => AUTH_ADMIN, 'Id' => 0, 'Role' => AUTH_ADMIN], 'Player', $mid, null, [
                 'authorization_id' => $authId,
                 'mundane_id'       => $mid,
                 'park_id'          => 0,
```
baltinerdist, line 66: **24. Grants are audited, revocations are not — the new reusable audit hop makes closing that gap nearly free**

💡 **medium · enhancement** · _effort: medium_

**What:** Every `addauth` branch this PR touched now logs an audit row through the new `Authorization->audit()` seam. The sibling revocation branches in the same five files (AdminAjax 'removeauth', EventAjax::auth 'removeauth', KingdomAjax 'removeauth', ParkAjax 'removeauth', Unit 'deleteauth') call `del_auth()`/`remove_unit_auth()` and never audit at all, so revoking admin/kingdom/park/event/unit authority leaves no forensic trail while the grant that created it does. The gap predates this PR, but `EventAjax::remove_staff` already demonstrates the exact pattern needed (fetch PriorState, audit with post_state null) two functions away, and the reusable hop this PR introduces makes the fix mechanical.

**Do:** In each 'removeauth'/'deleteauth' branch, capture the authorization row before deleting (look it up by AuthorizationId, or have del_auth return it) and call `$this->Authorization->audit('Authorization::RemoveAuthorization', [...], 'Player', $mid, $priorAuthRow, null)`, mirroring the EventAjax::remove_staff pattern, so grants and revocations are equally traceable.

<sub>Anchored to the nearest line inside this PR's diff; the finding refers to `orkui/controller/controller.AdminAjax.php:74-83; controller.KingdomAjax.php:607-620; controller.ParkAjax.php:392-405; controller.EventAjax.php:339-352; controller.Unit.php:171-176`.</sub>

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/ork3/class.AuthorizationGate.php
```diff
@@ -30,4 +30,14 @@ public function HasAuthority(array $request): array
             'Authorized' => $this->check($mundaneId, $type, $id, $role),
         ];
     }
+
+    /**
```
baltinerdist, line 34: **25. clearSharedDb relocates `global $DB; $DB->Clear()` without removing the coupling it was meant to remove**

🟡 **medium · problem** · _effort: small_

**What:** The headline claim is that class.Controller.php no longer touches raw `$DB`. In practice `global $DB; $DB->Clear();` moved verbatim into `AuthorizationGate::clearSharedDb()` and the controller now reaches it via three hops. The caller obligation — 'after you do auth checks you must remember to clear the shared connection' — is fully preserved, so any new auth-check caller that forgets reintroduces the leak. It also pushes a frontend workflow concept into a domain API name (`clear_db_after_auth_checks`) and leaves the `rg '\$DB->' orkui/` gate reporting green on a controller that still semantically mutates global DB state.

**Do:** Make the gate clean up after itself: call `$DB->Clear()` at the end of `AuthorizationGate::check()` (and `HasAuthority()`), then delete `Model_Authorization::clear_db_after_auth_checks()` and the call in the Controller constructor entirely. The invariant then holds for every caller, not just the one that remembered — which is the difference between a real layer boundary and a renamed global.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

docs/megiddo/refactor/idioms-00-charter.md
```diff
@@ -145,8 +154,10 @@ Run from repo root. **Exit 0** = pass for PHPUnit; **exit 1 (no matches)** = pas
 ### 4.1 Static isolation (every I-* hop)
 
 ```bash
-rg '\$DB->' orkui/                    # expect: no matches
-rg 'Ork3::\$Lib' orkui/              # expect: no matches
+rg '\$DB->' orkui/                                          # expect: no matches
```
baltinerdist, line 157: **26. The static gates that are supposed to prevent regression are four ripgrep lines in a markdown file with no executor**

💡 **medium · enhancement** · _effort: small_

**What:** The stated intent is that Lib access and inline Dangeraudit 'cannot creep back'. The mechanism shipped is idioms-00-charter.md §2/§4.1 listing four `rg` commands as 'non-negotiable'. Nothing runs them: there is no .github/workflows directory, and .githooks/pre-commit only unstages class.Authorization.php and runs php-cs-fixer. I confirmed all four greps do pass at this head SHA — the refactor's substance is real — but the enforcement is a convention someone has to remember, which is the exact failure mode the refactor exists to prevent. The Dangeraudit pattern is also over-narrow: it matches only the parenthesized `(new Dangeraudit())` form, so `$audit = new Dangeraudit();` or `new \Dangeraudit()` in a controller sails through.

**Do:** Add the four checks to .githooks/pre-commit (grepping only the staged file list so the hook stays fast) and/or a minimal CI workflow that fails on a match, and broaden the audit pattern to `new\s+\\?Dangeraudit\s*\(`. Consider a fifth gate asserting `orkui/model/model.*.php` contains no `->db->` / `DataSet(` / `Execute(` — the invariant that keeps the model layer thin, which is the property this whole PR is trying to establish.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

system/lib/ork3/class.Player.php
baltinerdist, line 369: **27. New aggregates trust a caller-supplied Awards array with no ownership check, ahead of their planned service registration**

💡 **medium · enhancement** · _effort: small_

**What:** `GetReconcilePageData`, `GetPlayerMilestones` and `GetLadderProgress` all follow the same pattern: if `Awards` is supplied use it as-is, otherwise validate MundaneId and fetch from the DB. When Awards is supplied, none of the three verifies that the rows belong to MundaneId. Every caller today is same-request controller code, so this is currently safe — but 02-api-contract.md §D explicitly proposes registering these exact methods as `PlayerService.*` JSON-RPC/SOAP entry points in a near-term follow-up. Registered as-is, an external caller could pass a real MundaneId alongside a fabricated Awards array and receive back milestones, ladder Rank tiles and reconcile RankSuggestions computed on data that never came from the DB — output identical in shape to the authoritative path and usable to justify a rank grant through the separate reconcile-commit endpoint.

**Do:** Before any orkservice/SOAP registration, either drop the trust-the-caller branch for externally-reachable entry points (always resolve Awards internally from MundaneId), or assert that every row's MundaneId, when present, matches the request's. Add a characterization test that passes a row with a mismatched MundaneId and asserts it is rejected or ignored, to lock the invariant in before the surface exists.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

orkui/template/revised-frontend/Playernew_index.tpl
```diff
@@ -129,11 +109,13 @@
...
-		9=>44, 10=>45, 11=>46, 12=>47, 14=>242, 15=>49, 16=>50, 17=>51,
-	];
+	// Class → Paragon award map from domain (controller-assigned ClassParagonMap)
+	if (!isset($ClassParagonMap) || !is_array($ClassParagonMap)) {
+		$ClassParagonMap = [];
+	}
+	if (!isset($ClassLevelThresholds) || !is_array($ClassLevelThresholds)) {
+		$ClassLevelThresholds = [];
+	}
 	$pnHeldAwardIds = [];
```
baltinerdist, line 119: **28. Template now computes held-award IDs twice, feeding a PnConfig key that is also defined twice**

💡 **low · enhancement** · _effort: trivial_

**What:** To rebuild the Paragon-badge wiring after the ladder block was deleted, the PR re-created `$pnHeldAwardIds` in the template preamble from `$Details['Awards']`. The file already computes `$playerHeldAwardIds` from the same source for award pre-fill/duplicate detection, and PnConfig ends up with two `heldAwardIds` keys where the second silently wins. The values match today because both filter AwardId > 0 over the same list, but that is two independently-maintained code paths for one client-side value — the drift risk the PR's own anti-pattern list calls out. (The duplicate PnConfig key itself predates this PR; the second computation is what the PR re-introduced.)

**Do:** Delete the new `$pnHeldAwardIds` block and reuse the existing `$playerHeldAwardIds` for both the Class Levels Paragon badge and the PnConfig emission (hoisting its computation above line 119 if needed), then drop the now-redundant duplicate `heldAwardIds` key so there is exactly one held-award set and one key.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

kenwalker:
Merged into master on 2026-08-18 as part of the 492+493+494 stack (fast-forward via local integration branch; this PR's head 2d007583 is reachable from master). Closing to reconcile status — content is fully shipped and live. Conflict resolution in orkui/model/model.Event.php is covered by tests/Unit/ModelEventMergeSeamTest.php + tools/infection/infection.t-merge-seam.json5.

Issue #23, opened 2014-06-09
[Park] Move/Merge field data loss


jrtaylor-com:
Old ref: B64

KingdomId is not getting set on Attendance, Mundane records

baltinerdist:
Closed with the 2026 UI revision (PR #416).

Issue #11, opened 2014-06-09
[Park] Waiver import from ORK2 didn't  happen


jrtaylor-com:
Old Ref: B20

Waivers did not transfer from ORK 2

esdraelon:
Almost four years old.  Not going to happen.

Issue #10, opened 2014-06-09
[Kingdom] Roles


jrtaylor-com:
Old Ref: B18

Looks like import bug, because China

esdraelon:
Insufficient information to be workable.

PR #494, opened 2026-07-20
Fuzzy Validator 2.1: containerized capture/validate runner

esdraelon:
## Summary
Stacked on [#493](https://github.com/amtgard/ORK3/pull/493) (`megiddo/p3-bootstrap-model-hop`). Merge **#492 → #493 → this PR** (or retarget after lower PRs land).

Makes the **default** `bin/fuzzy-validator` path run capture/validate inside a long-lived **Ubuntu 26.04** runner with Playwright-pinned Chromium, so setpoints are not host-OS/Chrome dependent.

- Auto-starts the runner if down; leaves it running after commands; `restart: "no"` (no reboot persistence).
- Joins the php8 Docker network (`http://ork3-php8-app/orkui/`); mounts the repo + `docker.sock` so dual-profile `ork-db` still works.
- Escape hatch: `FUZZY_VALIDATOR_NATIVE=1` / `--host` for bare-metal.
- Docs: `docs/megiddo/fuzzy-validator/version-2.1/` + tools README updates.

## Test plan
- [ ] `docker compose -f docker-compose.php8.yml up -d`
- [ ] `bin/fuzzy-validator validate --profile test --page home-authenticated --phase visual` (auto-starts runner)
- [ ] `docker stop ork3-fuzzy-validator-runner` then re-run a command — runner comes back
- [ ] `FUZZY_VALIDATOR_NATIVE=1 bin/fuzzy-validator validate --help` still works
- [ ] After merge: **re-record and publish gold setpoint from the runner** before treating visual gates as sign-off (existing macOS baselines will drift under Linux Chromium)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/megiddo/fuzzy-validator/README.md
+++ b/docs/megiddo/fuzzy-validator/README.md
@@ -4,7 +4,7 @@ Automated refactor-stability harness for Megiddo R-*: **hard CSS/JS byte checks*
 
 **Run:** `bin/fuzzy-validator record|validate|refuzz|setpoint …`  
 **Code:** `tools/fuzzy-validator/`  
-**Status:** v1 shipped (FU-0 … FU-16). Next product work: [version-2/](./version-2/).
+**Status:** v1 shipped (FU-0 … FU-16). v2 overlays shipped: [version-2/](./version-2/). Next product work (plan): [version-2.1/](./version-2.1/) — containerized Ubuntu 26.04 + pinned Chromium runner for stable setpoints.
 
 ---
 
@@ -38,8 +38,9 @@ open tools/fuzzy-validator/reports/run-*/index.html
 |------|----------|----------|
 | *(this folder)* | Humans | README, USER-GUIDE, DEVELOPER-GUIDE, design overview |
 | [reference/](./reference/) | Humans + agents | Live as-built specs (CLI, schemas, ops, reports, profiles) |
-| [skills/](./skills/) | Agents | Orchestration skills (reserved; empty until version-2) |
-| [version-2/](./version-2/) | Planners / agents | Home for the next feature plan (classified drift overlays) |
+| [skills/](./skills/) | Agents | Orchestration skills (run-setpoint-drift, putative-drift-overlay) |
+| [version-2/](./version-2/) | Planners / agents | Classified drift overlays (implemented) |
+| [version-2.1/](./version-2.1/) | Planners / agents | Containerized runner plan (Ubuntu 26.04 + pinned Chromium) — **status: plan** |
 | [archive/](./archive/) | Historical | Completed FU-* plans, prompts, checklists |
 
 ### Reorganization note (2026-07-19)
--- a/docs/megiddo/fuzzy-validator/README.md
+++ b/docs/megiddo/fuzzy-validator/README.md
@@ -4,7 +4,7 @@ Automated refactor-stability harness for Megiddo R-*: **hard CSS/JS byte checks*
 
 **Run:** `bin/fuzzy-validator record|validate|refuzz|setpoint …`  
 **Code:** `tools/fuzzy-validator/`  
-**Status:** v1 shipped (FU-0 … FU-16). v2 overlays shipped: [version-2/](./version-2/). Next product work (plan): [version-2.1/](./version-2.1/) — containerized Ubuntu 26.04 + pinned Chromium runner for stable setpoints.
+**Status:** v1 shipped (FU-0 … FU-16). v2 overlays shipped: [version-2/](./version-2/). **v2.1 runner shipped:** [version-2.1/](./version-2.1/) — default containerized Ubuntu 26.04 + pinned Chromium.
 
 ---
 
@@ -26,6 +26,7 @@ docker compose -f docker-compose.php8.yml up -d
 bin/ork-db deploy-sandbox
 bin/fuzzy-validator setpoint restore
 
+# Default path auto-starts the Ubuntu 26.04 runner (pinned Chromium).
 bin/fuzzy-validator validate --pages home-anonymous,player-profile --phase all
 open tools/fuzzy-validator/reports/run-*/index.html
 ```
--- a/docs/megiddo/fuzzy-validator/USER-GUIDE.md
+++ b/docs/megiddo/fuzzy-validator/USER-GUIDE.md
@@ -22,19 +22,22 @@ It produces **exit code pass/fail** (for CI) and an **HTML report** (for humans)
 
 ## One-time setup
 
+**Default path (recommended):** Docker + php8 stack. The CLI starts a long-lived **fuzzy-validator runner** (Ubuntu 26.04 + Playwright-pinned Chromium). Host Node/Python/Playwright installs are **not** required for day-to-day `bin/fuzzy-validator`.
+
 ```bash
 # Repo root
-npm ci
-npx playwright install chromium
-pip install -r tools/fuzzy-validator/python/requirements.txt
-
 docker compose -f docker-compose.php8.yml up -d
-bin/ork-db deploy-sandbox              # required for test profile
+bin/ork-db deploy-sandbox              # required for test profile (run on host)
 
+# Optional: host browsers / humans still use localhost
 export ORK3_E2E_BASE_URL=http://localhost:19080/orkui/
 export ORK3_E2E_USERNAME=admin ORK3_E2E_PASSWORD=password   # mirror (local docker)
 export ORK3_E2E_TEST_PASSWORD=test-db-player               # test profile (optional)
 ```
+
+First `bin/fuzzy-validator …` builds/starts `ork3-fuzzy-validator-runner` if needed (leaves it running; `restart: "no"`).
+
+**Native escape hatch** (debug only — not sign-off): `FUZZY_VALIDATOR_NATIVE=1` or `--host` (requires host `npm ci`, `npx playwright install chromium`, and pip deps).
 
 ### Restore baselines (required after clone)
 
@@ -272,7 +275,7 @@ Details: [04-operating-guide.md §10](./reference/04-operating-guide.md).
 | Record aborts (assets unstable) | Fix flaky scripts; stub network |
 | Empty fuzz manifest | Page fully deterministic — OK |
 | Dimension mismatch | Content height changed — update baseline |
-| macOS local fail, Linux pass | Prefer a Linux docker/host gate as the sign-off surface when screenshots disagree |
+| macOS native vs runner visual disagree | Expected if baselines were recorded on host Chrome — re-record / publish setpoint via default (container) path; do not use `--host` for gold master |
 
 More: [04-operating-guide.md §7](./reference/04-operating-guide.md).
 
--- a/tools/fuzzy-validator/README.md
+++ b/tools/fuzzy-validator/README.md
@@ -752,20 +752,34 @@ ORK3 uses this tool to gate front-end refactors against local docker. Two profil
 
 ### Quick start (ORK3)
 
+**Default:** containerized runner (Ubuntu 26.04 + Playwright-pinned Chromium). Host Playwright/Python installs are optional.
+
 ```bash
-npm ci
-npx playwright install chromium
-pip install -r tools/fuzzy-validator/python/requirements.txt
-
 docker compose -f docker-compose.php8.yml up -d
-bin/ork-db deploy-sandbox --yes
+bin/ork-db deploy-sandbox
 
-export ORK3_E2E_BASE_URL=http://localhost:19080/orkui/
+# Humans / host browsers still use localhost; the runner defaults to
+# http://ork3-php8-app/orkui/ on the php8 docker network.
 export ORK3_E2E_USERNAME=admin ORK3_E2E_PASSWORD=password
 # test profile defaults: megiddo / test-db-player
 
 bin/fuzzy-validator setpoint restore
 bin/fuzzy-validator validate --pages home-authenticated,player-profile --phase all
+```
+
+First invocation builds/starts `ork3-fuzzy-validator-runner` if needed and **leaves it running** (`restart: "no"` — no reboot persistence).
+
+```bash
+# Stop / rebuild the runner when the Playwright pin or Dockerfile changes:
+docker compose -f tools/fuzzy-validator/docker-compose.runner.yml stop
+FUZZY_VALIDATOR_REBUILD=1 bin/fuzzy-validator validate --help   # or: compose up -d --build
+
+# Non-default compose project network:
+export FUZZY_VALIDATOR_DOCKER_NETWORK=myproject_ork3-php8-net
+
+# Native escape hatch (debug only — captures may differ from sign-off):
+FUZZY_VALIDATOR_NATIVE=1 bin/fuzzy-validator validate --page home-authenticated
+# or: bin/fuzzy-validator --host validate …
 ```
 
 Open `tools/fuzzy-validator/reports/run-*/index.html`.
@@ -790,11 +804,13 @@ Default `validate` / `record` run both profiles unless you pass `--profile` / `-
 
 - Run `setpoint restore` after clone; otherwise `validate` exits `2`.  
 - Prefer `bin/fuzzy-validator` over legacy `calibrate.sh` / `gate.sh`.  
-- Screenshot compare can differ macOS vs Linux; prefer a **Linux** docker/host gate for sign-off when they disagree. **Planned default:** containerized runner (Ubuntu 26.04 + pinned Chromium) so capture/validate are stable across host OS — see [`docs/megiddo/fuzzy-validator/version-2.1/`](../../docs/megiddo/fuzzy-validator/version-2.1/) (**status: plan**).  
+- **Sign-off path is the containerized runner** ([version-2.1](../../docs/megiddo/fuzzy-validator/version-2.1/)). After switching to 2.1, **re-record and publish the gold setpoint once** from the runner — macOS-native baselines often fail visual under Linux Chromium with no product change.  
+- Cloud-synced worktrees (e.g. Google Drive bind mounts) can be slow/flaky for the runner volume; prefer a local disk clone for heavy loops.  
 - Do not lower `assetsMinScore` to hide a refactor; fix the drift or intentionally re-record.  
 - Do not use overlays (or annotations) to soft-pass unexpected regressions; overlays are for *declared* expected drift only.  
 - `validate --all` is slow (many pages × two profiles).  
 - For dual-profile sign-off with a freshness check: `validate … --require-fresh-mirror` (override only with `--mirror-stale-ok "reason"`). Refresh the local prod mirror / re-`bin/ork-db extract` when prompted.
+- Native mode (`--host` / `FUZZY_VALIDATOR_NATIVE=1`) is for tool debugging only — not gold-master capture.
 
 Agent orchestration skills (run setpoint drift; draft putative overlays from requirements): [`docs/megiddo/fuzzy-validator/skills/`](../../docs/megiddo/fuzzy-validator/skills/).
 
--- a/tools/fuzzy-validator/README.md
+++ b/tools/fuzzy-validator/README.md
@@ -790,7 +790,7 @@ Default `validate` / `record` run both profiles unless you pass `--profile` / `-
 
 - Run `setpoint restore` after clone; otherwise `validate` exits `2`.  
 - Prefer `bin/fuzzy-validator` over legacy `calibrate.sh` / `gate.sh`.  
-- Screenshot compare can differ macOS vs Linux; prefer a **Linux** docker/host gate for sign-off when they disagree.  
+- Screenshot compare can differ macOS vs Linux; prefer a **Linux** docker/host gate for sign-off when they disagree. **Planned default:** containerized runner (Ubuntu 26.04 + pinned Chromium) so capture/validate are stable across host OS — see [`docs/megiddo/fuzzy-validator/version-2.1/`](../../docs/megiddo/fuzzy-validator/version-2.1/) (**status: plan**).  
 - Do not lower `assetsMinScore` to hide a refactor; fix the drift or intentionally re-record.  
 - Do not use overlays (or annotations) to soft-pass unexpected regressions; overlays are for *declared* expected drift only.  
 - `validate --all` is slow (many pages × two profiles).  
--- a/tools/fuzzy-validator/bin/fuzzy-validator
+++ b/tools/fuzzy-validator/bin/fuzzy-validator
@@ -1,9 +1,15 @@
 #!/usr/bin/env bash
-# Fuzzy Validator — CLI dispatcher
+# Fuzzy Validator — CLI dispatcher (default: containerized runner).
+#
+# Default path: ensure ork3-fuzzy-validator-runner is up, then docker exec.
+# Escape hatch: FUZZY_VALIDATOR_NATIVE=1 or --host → bare-metal Python/Playwright.
 set -euo pipefail
 
 TOOL_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
-PYTHON_DIR="$TOOL_ROOT/python"
+REPO_ROOT="$(cd "$TOOL_ROOT/../.." && pwd)"
+COMPOSE_FILE="$TOOL_ROOT/docker-compose.runner.yml"
+RUNNER_NAME="ork3-fuzzy-validator-runner"
+NATIVE_BIN="$TOOL_ROOT/bin/fuzzy-validator-native"
 
 usage() {
   cat <<'EOF'
@@ -22,31 +28,149 @@ Commands:
 
 Global options:
   -h, --help    Show help
+  --host        Run on the host (native Playwright); not the sign-off path
+
+Environment:
+  FUZZY_VALIDATOR_NATIVE=1           Same as --host
+  FUZZY_VALIDATOR_DOCKER_NETWORK=…   Override php8 external network name
+  FUZZY_VALIDATOR_REBUILD=1          Force runner image rebuild on ensure
+  ORK3_E2E_*                         Passed through to the runner
+
+Default execution is inside the Ubuntu 26.04 runner container (pinned Chromium).
+See tools/fuzzy-validator/README.md and docs/megiddo/fuzzy-validator/version-2.1/.
 
 Run from repo root:
   bin/fuzzy-validator record --help
   bin/fuzzy-validator validate --help
-
-Documentation:
-  docs/megiddo/fuzzy-validator/10-cli-reference.md
 EOF
 }
 
-cmd="${1:-}"
-shift || true
+# Already inside the runner → never nest docker exec.
+if [[ "${FUZZY_VALIDATOR_IN_CONTAINER:-}" == "1" ]]; then
+  exec "$NATIVE_BIN" "$@"
+fi
 
-case "$cmd" in
-  record|validate|refuzz|setpoint|overlay)
-    export PYTHONPATH="${PYTHONPATH:-}:$PYTHON_DIR"
-    exec python3 -m fuzzy_validator.cli "$cmd" "$@"
-    ;;
-  -h|--help|help|'')
-    usage
-    exit 0
-    ;;
-  *)
-    echo "fuzzy-validator: unknown command '$cmd'" >&2
-    usage
-    exit 2
-    ;;
-esac
+want_native=0
+if [[ "${FUZZY_VALIDATOR_NATIVE:-}" == "1" ]]; then
+  want_native=1
+fi
+
+forward_args=()
+for arg in "$@"; do
+  if [[ "$arg" == "--host" ]]; then
+    want_native=1
+    continue
+  fi
+  forward_args+=("$arg")
+done
+
+if [[ "$want_native" -eq 1 ]]; then
+  echo "fuzzy-validator: native mode — captures may differ from containerized default" >&2
+  exec "$NATIVE_BIN" "${forward_args[@]}"
+fi
+
+# Top-level help without docker: avoid forcing a build for bare --help.
+cmd="${forward_args[0]:-}"
+if [[ -z "$cmd" || "$cmd" == "-h" || "$cmd" == "--help" || "$cmd" == "help" ]]; then
+  usage
+  exit 0
+fi
+
+detect_php8_network() {
+  if [[ -n "${FUZZY_VALIDATOR_DOCKER_NETWORK:-}" ]]; then
+    printf '%s\n' "$FUZZY_VALIDATOR_DOCKER_NETWORK"
+    return 0
+  fi
+  if ! docker inspect ork3-php8-app >/dev/null 2>&1; then
+    echo "fuzzy-validator: ork3-php8-app is not running." >&2
+    echo "Start the php8 stack first: docker compose -f docker-compose.php8.yml up -d" >&2
+    return 1
+  fi
+  local nets first
+  nets="$(docker inspect ork3-php8-app -f '{{range $k, $v := .NetworkSettings.Networks}}{{println $k}}{{end}}' 2>/dev/null || true)"
+  first="$(printf '%s\n' "$nets" | grep -E 'php8-net' | head -n1 || true)"
+  if [[ -z "$first" ]]; then
+    first="$(printf '%s\n' "$nets" | head -n1 | tr -d '[:space:]')"
+  fi
+  if [[ -z "$first" ]]; then
+    echo "fuzzy-validator: could not detect docker network for ork3-php8-app" >&2
+    echo "Set FUZZY_VALIDATOR_DOCKER_NETWORK explicitly." >&2
+    return 1
+  fi
+  printf '%s\n' "$first"
+}
+
+runner_running() {
+  local state
+  state="$(docker inspect -f '{{.State.Running}}' "$RUNNER_NAME" 2>/dev/null || echo false)"
+  [[ "$state" == "true" ]]
+}
+
+ensure_runner_running() {
+  local net
+  net="$(detect_php8_network)" || return 1
+  export FUZZY_VALIDATOR_DOCKER_NETWORK="$net"
+
+  if runner_running; then
+    return 0
+  fi
+
+  echo "fuzzy-validator: starting runner (network=$net)…" >&2
+
+  local need_build=0
+  if [[ "${FUZZY_VALIDATOR_REBUILD:-}" == "1" ]] || ! docker image inspect ork3-fuzzy-validator-runner:local >/dev/null 2>&1; then
+    need_build=1
+  fi
+
+  (
+    cd "$REPO_ROOT"
+    if [[ "$need_build" -eq 1 ]]; then
+      FUZZY_VALIDATOR_DOCKER_NETWORK="$net" \
+        docker compose -f "$COMPOSE_FILE" up -d --build
+    else
+      FUZZY_VALIDATOR_DOCKER_NETWORK="$net" \
+        docker compose -f "$COMPOSE_FILE" up -d
+    fi
+  )
+
+  local i
+  for i in $(seq 1 90); do
+    if docker exec "$RUNNER_NAME" true >/dev/null 2>&1; then
+      return 0
+    fi
+    sleep 1
+  done
+  echo "fuzzy-validator: runner did not become ready (container=$RUNNER_NAME)" >&2
+  return 1
+}
+
+ensure_runner_running || exit 1
+
+exec_env=(
+  -e FUZZY_VALIDATOR_IN_CONTAINER=1
+  -e "ORK3_E2E_BASE_URL=${ORK3_E2E_BASE_URL:-http://ork3-php8-app/orkui/}"
+  -e PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
+)
+
+# Pass through auth, clock, CI, and fuzzy/fuzz knobs when set on the host.
+pass_keys=(
+  ORK3_E2E_USERNAME
+  ORK3_E2E_PASSWORD
+  ORK3_E2E_PLAYER_ID
+  ORK3_CLOCK_DATE
+  CI
+  FUZZY_VALIDATOR_SKIP_NPM_CI
+  FUZZ_PAGES
+  FUZZ_MODE
+  FUZZ_REPEAT
+  FUZZ_TOOL_ROOT
+  FUZZ_PAGES_MANIFEST
+)
+for key in "${pass_keys[@]}"; do
+  if [[ -n "${!key:-}" ]]; then
+    exec_env+=(-e "${key}=${!key}")
+  fi
+done
+
+exec docker exec -i -w /ork3 "${exec_env[@]}" "$RUNNER_NAME" \
+  /ork3/tools/fuzzy-validator/bin/fuzzy-validator-native "${forward_args[@]}"
--- a/tools/fuzzy-validator/docker-compose.runner.yml
+++ b/tools/fuzzy-validator/docker-compose.runner.yml
@@ -1,0 +1,41 @@
+# Fuzzy Validator 2.1 — long-lived capture/validate runner (not part of php8 stack).
+#
+# Network: joins the external php8 user network. Default name matches compose
+# project "ork3"; override with FUZZY_VALIDATOR_DOCKER_NETWORK or let the CLI
+# wrapper auto-detect from `ork3-php8-app`.
+#
+# Lifecycle: restart "no" — auto-started by CLI; left running; does not survive reboot.
+
+services:
+  fuzzy-validator-runner:
+    build:
+      context: ../..
+      dockerfile: tools/fuzzy-validator/docker/Dockerfile
+    image: ork3-fuzzy-validator-runner:local
+    container_name: ork3-fuzzy-validator-runner
+    restart: "no"
+    working_dir: /ork3
+    volumes:
+      - ../..:/ork3
+      - fv-node-modules:/ork3/node_modules
+      - /var/run/docker.sock:/var/run/docker.sock
+    environment:
+      ORK3_E2E_BASE_URL: ${ORK3_E2E_BASE_URL:-http://ork3-php8-app/orkui/}
+      FUZZY_VALIDATOR_IN_CONTAINER: "1"
+      PLAYWRIGHT_BROWSERS_PATH: /ms-playwright
+      # Host docker.sock paths in compose files refer to the host engine.
+      DOCKER_HOST: unix:///var/run/docker.sock
+    networks:
+      - ork3-php8-net
+    stdin_open: true
+    tty: true
+    command: ["sleep", "infinity"]
+
+volumes:
+  fv-node-modules:
+    name: ork3-fv-node-modules
+
+networks:
+  ork3-php8-net:
+    external: true
+    name: ${FUZZY_VALIDATOR_DOCKER_NETWORK:-ork3_ork3-php8-net}
--- a/tools/fuzzy-validator/docker/Dockerfile
+++ b/tools/fuzzy-validator/docker/Dockerfile
@@ -1,0 +1,68 @@
+# Fuzzy Validator 2.1 runner — Ubuntu 26.04 + Playwright-pinned Chromium.
+#
+# node_modules strategy:
+#   Named volume `fv-node-modules` overlays /ork3/node_modules so Linux
+#   native addons never clash with a macOS host tree. Entrypoint runs
+#   `npm ci` into that volume when @playwright/test is missing.
+#   Playwright browser binaries live in-image at PLAYWRIGHT_BROWSERS_PATH
+#   (not the host ~/Library/Caches/ms-playwright).
+#
+# Pin: ubuntu:resolute-20260707 (== ubuntu:26.04 at spike time).
+
+ARG UBUNTU_IMAGE=ubuntu:resolute-20260707
+FROM ${UBUNTU_IMAGE}
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    LANG=C.UTF-8 \
+    PLAYWRIGHT_BROWSERS_PATH=/ms-playwright \
+    FUZZY_VALIDATOR_IN_CONTAINER=1 \
+    PATH=/opt/fv-venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
+# OS packages: Playwright Chromium deps installed via `playwright install --with-deps`.
+# Also: Node 22, Python 3 + venv, PHP CLI (ork-db), Docker CLI (compose restart / exec).
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+        wget \
+        git \
+        python3 \
+        python3-pip \
+        python3-venv \
+        nodejs \
+        npm \
+        php8.5-cli \
+        php8.5-mysql \
+        php8.5-mbstring \
+        php8.5-xml \
+        php8.5-curl \
+        docker.io \
+        docker-compose-v2 \
+        socat \
+        fonts-liberation \
+        fonts-dejavu-core \
+        fontconfig \
+    && rm -rf /var/lib/apt/lists/*
+
+# Python tool deps (baked venv; repo mount does not need host pip).
+COPY tools/fuzzy-validator/python/requirements.txt /tmp/fv-requirements.txt
+RUN python3 -m venv /opt/fv-venv \
+    && /opt/fv-venv/bin/pip install --no-cache-dir -U pip \
+    && /opt/fv-venv/bin/pip install --no-cache-dir -r /tmp/fv-requirements.txt \
+    && rm -f /tmp/fv-requirements.txt
+
+# Playwright / Chromium pinned to repo lockfile version (install browsers into image).
+WORKDIR /tmp/fv-playwright-pin
+COPY package.json package-lock.json ./
+RUN npm ci --ignore-scripts \
+    && npx playwright install --with-deps chromium \
+    && node -e "const p=require('playwright/package.json'); const fs=require('fs'); const {execSync}=require('child_process'); let chromium='unknown'; try { chromium=execSync('ls /ms-playwright').toString().trim().split('\\n').filter(Boolean).join(', '); } catch (e) {} fs.writeFileSync('/opt/BROWSER_PIN.txt', ['playwright=' + p.version, 'playwright_browsers_path=/ms-playwright', 'chromium_cache=' + chromium, 'node=' + process.version, 'recorded_at=' + new Date().toISOString(), ''].join('\\n'));" \
+    && rm -rf /tmp/fv-playwright-pin
+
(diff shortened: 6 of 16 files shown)


baltinerdist (commented):
## Adversarial review — ship with fixes

Reviewed stacked on #493 + #492-with-review-fixes (a hostile multi-agent panel: attack premises across the CLI, Dockerfile, entrypoint and compose file, each finding then attacked by an independent refuter — one of the raised findings died there).

**The plumbing is right.** Verified by reading: `detect_php8_network()` auto-detects and exports the real project-derived network name before compose runs, correctly masking the hardcoded `ork3-php8-net` fallback in `docker-compose.runner.yml`; the socat DB forward targets (`ork3-php8-db`, `ork3-php8-test-db`) match `docker-compose.php8.yml`'s `container_name`s; the non-`exec` entrypoint is deliberate and correct, to keep the socat children alive under PID 1; `--ignore-scripts` is on **both** the build-time and runtime `npm ci` with no path that omits it; the `fv-node-modules` named volume shadows only `/ork3/node_modules` rather than the whole bind mount; `fuzzy-validator-native` is a faithful extraction of the old dispatcher (same command whitelist, `PYTHONPATH` export, exit-2 on unknown command); and `FUZZY_VALIDATOR_IN_CONTAINER=1` correctly blocks a nested `docker exec`.

Two defects in `ensure_runner_running`, both of which cause the tool to produce a **silently wrong verdict** — the exact failure class this PR exists to eliminate.

---

### 🟠 High — the runner is reused on name alone, without checking which repo its `/ork3` bind mount points at

`tools/fuzzy-validator/bin/fuzzy-validator:103-116`, with `../..:/ork3` declared at `docker-compose.runner.yml:19`.

```bash
runner_running() {
  local state
  state="$(docker inspect -f '{{.State.Running}}' "$RUNNER_NAME" 2>/dev/null || echo false)"
  [[ "$state" == "true" ]]
}

ensure_runner_running() {
  ...
  if runner_running; then
    return 0            # <-- never reaches `docker compose up -d`
  fi
```

**Trigger:** two checkouts of ORK3 on one host — the normal state for anyone using git worktrees, and the normal state for anyone reviewing this stack.

Run `bin/fuzzy-validator validate` from clone A. The runner is created with `/ork3` bound to A and, by design (`restart: "no"`, left running after commands), stays up. Now `cd` to worktree B and run `validate`, `record`, or `setpoint publish`. `ensure_runner_running` computes `REPO_ROOT` from the invoked script's own path — correctly, B — but the only reuse check is `runner_running`, which asks docker for `.State.Running` on the fixed name `ork3-fuzzy-validator-runner` and returns 0 at :115. Nothing compares the container's `/ork3` mount `Source` to `REPO_ROOT`, and because it returns early it never reaches `docker compose up -d`, which is the one call that would have noticed the changed bind mount and recreated the container.

The final `docker exec -i -w /ork3` then runs B's command line entirely inside **A's** tree.

**Consequence:** `validate` prints PASS/FAIL computed from clone A's templates against A's baselines while the developer believes it gated B. `record` / `refuzz` / `setpoint publish` are worse — they write new baselines, fuzz manifests and `run-*` reports into clone A's `tools/fuzzy-validator/baselines` and `reports`, mutating a working tree the developer never touched and did not expect to be written to.

**Suggested fix:** before the early return, compare the actual mount:

```bash
mounted="$(docker inspect -f '{{range .Mounts}}{{if eq .Destination "/ork3"}}{{.Source}}{{end}}{{end}}' "$RUNNER_NAME" 2>/dev/null)"
if runner_running && [[ "$mounted" == "$REPO_ROOT" ]]; then
  return 0
fi
```

…and otherwise fall through to `docker compose -f "$COMPOSE_FILE" up -d --force-recreate`, or hard-fail with an explicit `runner is bound to <mounted>, not <REPO_ROOT>`. Name-only identity is not sufficient for a container that carries a bind mount.

---

### 🟡 Medium — `FUZZY_VALIDATOR_REBUILD` is discarded when the runner is up, and a stale image is reused when it is down

`tools/fuzzy-validator/bin/fuzzy-validator:114-133` — the `if runner_running; then return 0; fi` at :114 sits **above** the `need_build` evaluation at :120-123.

```bash
  if runner_running; then
    return 0
  fi
  ...
  local need_build=0
  if [[ "${FUZZY_VALIDATOR_REBUILD:-}" == "1" ]] || ! docker image inspect ork3-fuzzy-validator-runner:local >/dev/null 2>&1; then
    need_build=1
  fi
```

**Path 1:** bump the Playwright pin in `package-lock.json` (or edit `docker/Dockerfile`), then run `FUZZY_VALIDATOR_REBUILD=1 bin/fuzzy-validator validate` with the runner still up — the steady state this PR deliberately creates. The early return short-circuits before `need_build` is computed, so the env var is silently dropped and no rebuild happens.

**Path 2:** `docker stop` the runner first and run without the env var. `need_build` is only set when `docker image inspect ork3-fuzzy-validator-runner:local` **fails**, and the previously built image still exists — so the script runs a plain `up -d` and starts the stale image.

The only two paths that ever rebuild are "image entirely absent" or "container stopped **and** `REBUILD=1`.

**Consequence:** capture keeps running the previously baked Chromium while `docker/BROWSER_PIN.txt` and `package-lock.json` advertise a newer one, with no warning emitted — a stale, host-artifact-contaminated verdict. It also undercuts the FV21-5 migration guidance ("re-record and publish the gold setpoint from the runner"), since the re-record can be produced by the old image.

**Suggested fix:** move the `need_build` evaluation above the running check, and when `FUZZY_VALIDATOR_REBUILD=1` run `docker compose -f "$COMPOSE_FILE" up -d --build --force-recreate` regardless of `runner_running`. Better still, on every ensure compare the running container's baked `/BROWSER_PIN.txt` (from the Dockerfile's `cp /opt/BROWSER_PIN.txt /BROWSER_PIN.txt`) against the repo's `tools/fuzzy-validator/docker/BROWSER_PIN.txt`, and rebuild or hard-fail on mismatch. That makes pin drift structurally unable to be validated silently, which is the whole point of the version-2.1 requirements doc.

---

### ⚠️ Unreviewed: the docker socket mount

`docker-compose.runner.yml:21` bind-mounts `/var/run/docker.sock` into a container that also runs headless Chromium against local pages. A reviewer raised that as critical; it was dropped from the panel's adjudication on a formatting technicality and **no refuter ever evaluated it**, so I'm flagging it as an open question rather than a finding.

The socket is root-equivalent control of the host daemon. The design doc justifies it so the dual-profile `ork-db` path keeps working — worth stating explicitly in `version-2.1/02-design.md` what inside the container is allowed to reach it, and whether any capture/validate code path shells out to `docker` (as opposed to only the `ork-db` helper). If only the DB helper needs it, scoping it — a socket proxy, or dropping the mount from the default profile and requiring an opt-in for `ork-db` — would be worth the trade.

### Coverage caveats

Both findings above are code-verified but **not empirically reproduced** — the runner was never actually started from a second worktree, and no pin bump was attempted. The round-2 pass over this shard died on an API error, so this PR got one attack pass rather than two.

baltinerdist (commented):
## Distributed Review — PR #494

PR #494's premise — capture/validate stops depending on the host OS and browser — is sound and the containerization is mostly well-executed (network auto-detect, node_modules volume isolation, `--ignore-scripts` to protect host assets, a working native escape hatch). But nothing in the shipped code actually *enforces* the pin or the identity of the runner, and that is where the critical findings cluster. I verified every claim below against the pristine checkout at c558cd3f; the 36 panel findings collapsed to 24 items after deduping (four lenses independently found the readiness race, three found the stale-image problem, three found the root/ownership problem).

Four issues are critical and I would fix them before merge. (1) Runner reuse is keyed only on the container name, and the container name, image tag, and node_modules volume are all host-global — so with the stacked worktrees this team actually uses, a runner started from checkout A will silently service commands issued from checkout B, running A's tool code against A's manifests and writing artifacts into A's tree, with no error. That is a silently-wrong gate result. (2) Image reuse checks only that the `:local` tag exists; nothing compares the Dockerfile, package-lock, or BROWSER_PIN against the built image, so a teammate can validate for weeks against a stale Chromium — and `docker/BROWSER_PIN.txt` is inert decoration (the Dockerfile never reads it, nothing compares it to the in-image copy, and the committed file has already drifted in shape from what the generator emits). (3) The readiness probe polls `docker exec … true`, which succeeds the instant the container starts — while PID 1 is still running a multi-minute `npm ci` and has not yet started the socat DB forwards — so the first-ever command races the entrypoint. (4) The USER-GUIDE setup block still exports `ORK3_E2E_BASE_URL=http://localhost:19080/orkui/`, which the wrapper forwards verbatim into the runner where localhost is the runner itself; the tool README correctly deleted that export, so the two docs now contradict each other and following the guide breaks the default path.

A fifth critical is architectural rather than a bug: the runner bind-mounts `/var/run/docker.sock` and runs as root, so any code execution inside it (Chromium rendering the app under test, an npm/pip dependency, the Python pipeline) is host-root. The stated need is narrow — letting `ork-db deploy-sandbox` run from inside the runner — and is better served by running that step on the host or fronting the socket with a scoped proxy.

Two gaps undercut the PR's own acceptance criterion: there is no `platform:` pin, so Apple Silicon and Linux x86 developers still build different Chromium binaries (the design doc decided linux/amd64 but nothing encodes it, and the FV21-5 parity evidence is arm64-only), and the base image is a mutable tag with floating font/fontconfig versions. Finally, the author's own re-record caveat lives only in prose: setpoints carry no environment metadata, so post-merge the stale macOS baselines will report as ~39 product regressions on every developer's first run — worth stamping a browser/arch/in-container block into captures and failing with "re-record from the runner" instead.


**24 findings** across 6 themes, from a panel of specialist reviewers (technical + subject-matter). 23 are attached inline below; 1 are cross-cutting and listed here.

Themes: Runner identity, reuse, and image freshness, Startup readiness and failure diagnostics, Container privileges and security, Documentation drift against the new default path, Cross-host determinism and setpoint provenance, Runner lifecycle ergonomics

### Cross-cutting findings (no single line to anchor to)

**21. Record the capture environment in each setpoint and fail loudly on mismatch, instead of relying on a README sentence** — 💡 high · enhancement · _effort: medium_
- **What:** The PR's own migration story is prose only: README.md:807 and USER-GUIDE.md:278 tell maintainers to re-record the gold setpoint once, and version-2.1/02-design.md:277 says "do not silently soft-fail", but nothing enforces it. `setpoint.json` carries only gitSha, capturedAt, contentSha256, profiles, pageCount — no browser build, Playwright version, image id, arch, viewport/DPR, or in-container flag; `FUZZY_VALIDATOR_IN_CONTAINER` is diagnostic only and is never persisted with artifacts. Two consequences: after merge the existing macOS baselines are indistinguishable from valid ones, so the first validate on every machine reports what FV21-5 measured (visual=0.957, 39 unexpected drifts) as if it were a product regression — training people to ignore red; and nothing stops `--host` from running `record` → `setpoint capture` → `setpoint publish`, so a host-captured gold master can be published behind only the one-line stderr notice at bin/fuzzy-validator:68, with no downstream way to tell.
- **Do:** Write an `environment` block alongside each captured baseline and into the setpoint bundle metadata: runner image id + build key, contents of `/BROWSER_PIN.txt`, arch, `FUZZY_VALIDATOR_IN_CONTAINER`, viewport/DPR, TZ/locale. Have `validate` compare it against the live environment and exit 2 with "baseline captured on host/macOS Chromium — re-record from the runner" rather than reporting product drift. Refuse `setpoint capture`/`publish` in native mode unless an explicit `--allow-native-setpoint` override is passed. As a stopgap for the merge itself, add the marker to the currently published setpoint so the first post-merge run explains itself.
- **Where:** `tools/fuzzy-validator/README.md:600-615; tools/fuzzy-validator/bin/fuzzy-validator:67-70`

---

Review performed by Claude, take with a grain of salt. -TofH

tools/fuzzy-validator/bin/fuzzy-validator
```diff
@@ -22,31 +28,149 @@ Commands:
...
+  fi
+  if [[ -z "$first" ]]; then
+    echo "fuzzy-validator: could not detect docker network for ork3-php8-app" >&2
+    echo "Set FUZZY_VALIDATOR_DOCKER_NETWORK explicitly." >&2
+    return 1
+  fi
+  printf '%s\n' "$first"
+}
+
+runner_running() {
```
baltinerdist, line 103: **1. Runner reuse is keyed only on container name — a runner started from another checkout silently services this one**

🔴 **critical · problem** · _effort: medium_

**What:** `ensure_runner_running()` treats the runner as usable whenever a container named `ork3-fuzzy-validator-runner` reports `.State.Running == true` (bin/fuzzy-validator:103-116). It never checks that the container's `/ork3` bind mount points at `$REPO_ROOT`, that it is on the network just detected, or that its image is the current one. `container_name`, the image tag `ork3-fuzzy-validator-runner:local`, and the volume `ork3-fv-node-modules` are all host-global constants, so exactly one runner can exist per machine. This repo is routinely worked in stacked worktrees (this review runs from `wt494` on top of `wt493`/`wt492`): run the CLI from checkout B while a runner started from checkout A is up and `docker exec -w /ork3 …` executes checkout A's tool code against checkout A's manifests, writing baselines/reports/setpoints into checkout A's tree — with no error and no warning. A silently wrong gate result is the worst failure mode a visual-regression harness can have.

**Do:** In `ensure_runner_running`, inspect the live container and compare the `Mounts` source for destination `/ork3` against `$REPO_ROOT` (and the joined network against the detected `$net`); on mismatch `docker rm -f` and recreate rather than exec. Better, scope the names per checkout: derive a short hash of `$REPO_ROOT` (or set `COMPOSE_PROJECT_NAME` from it) and suffix `container_name`, the image tag, and the `fv-node-modules` volume name so concurrent worktrees each get their own runner. Either way, echo the resolved repo path and container id once per invocation so a wrong-tree run is visible.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
```diff
@@ -22,31 +28,149 @@ Commands:
...
+ensure_runner_running() {
+  local net
+  net="$(detect_php8_network)" || return 1
+  export FUZZY_VALIDATOR_DOCKER_NETWORK="$net"
+
+  if runner_running; then
+    return 0
+  fi
+
+  echo "fuzzy-validator: starting runner (network=$net)…" >&2
```
baltinerdist, line 118: **2. Image reuse checks only that the tag exists, never that it matches the current Dockerfile / BROWSER_PIN / lockfile**

🔴 **critical · problem** · _effort: medium_

**What:** `need_build` is set only when `FUZZY_VALIDATOR_REBUILD=1` or the `ork3-fuzzy-validator-runner:local` tag is entirely absent (bin/fuzzy-validator:121). Nothing hashes the Dockerfile, `package-lock.json`, or `docker/BROWSER_PIN.txt` against the built image. Pull a branch that bumps the Playwright pin or edits the Dockerfile and every developer keeps validating against their months-old image and a different Chromium — silently, which directly defeats the PR's stated goal. `docker/BROWSER_PIN.txt` compounds this: it is committed as if it were the pin source, but the Dockerfile never reads it (the real pin is `ARG UBUNTU_IMAGE` + the repo lockfile), it is only *generated into* the image at `/BROWSER_PIN.txt` (Dockerfile:59,64), and nothing ever compares the two. The committed copy has already drifted in shape from what the generator emits — it carries `base_image=` and `note=` lines the build script does not produce, so it is hand-maintained and unverifiable.

**Do:** Bake a build key into the image — `LABEL fv.source_hash=` holding a hash of {Dockerfile, package-lock.json, BROWSER_PIN.txt} passed as a build-arg — and compare it in `ensure_runner_running` (before the `runner_running` short-circuit); on mismatch, print why and force a rebuild + container recreate instead of reusing. Additionally read `/BROWSER_PIN.txt` from the live runner and compare to the committed file, refusing sign-off commands (`record`, `setpoint capture`) on mismatch. Regenerate the committed `docker/BROWSER_PIN.txt` from the same code path that writes the in-image file so the two cannot diverge.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
```diff
@@ -22,31 +28,149 @@ Commands:
...
+  printf '%s\n' "$first"
+}
+
+runner_running() {
+  local state
+  state="$(docker inspect -f '{{.State.Running}}' "$RUNNER_NAME" 2>/dev/null || echo false)"
+  [[ "$state" == "true" ]]
+}
+
+ensure_runner_running() {
```
baltinerdist, line 109: **3. FUZZY_VALIDATOR_REBUILD=1 is silently ignored whenever the runner is already running**

🟠 **high · problem** · _effort: trivial_

**What:** `usage()` documents `FUZZY_VALIDATOR_REBUILD=1` as "Force runner image rebuild on ensure" with no caveat (bin/fuzzy-validator:36), but `ensure_runner_running()` returns at `if runner_running; then return 0; fi` (lines 114-116) before it ever reaches the `need_build` check that reads the variable (line 121). Since the runner is designed to be left running across invocations, that is the common case — and this flag is the only escape hatch from finding #2. The tool README's own rebuild recipe (README.md:773-775) only works because it happens to prepend a `compose … stop`; that ordering is load-bearing but unexplained.

**Do:** Evaluate `FUZZY_VALIDATOR_REBUILD` before the `runner_running` short-circuit; when set, stop and recreate the runner (or run `compose up -d --build`, which is safe against a running container) so the flag does what its own help text claims. Add the ordering note to the README recipe until then.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/docker/entrypoint.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+# Fuzzy-validator runner entrypoint: ensure Linux node_modules, forward
+# host-published DB ports to php8 containers, then idle or exec.
+set -euo pipefail
+
+REPO_ROOT="${FUZZY_VALIDATOR_REPO_ROOT:-/ork3}"
+cd "$REPO_ROOT"
+
+ensure_node_modules() {
```
baltinerdist, line 9: **4. `ensure_node_modules` only checks that `@playwright/test` exists, never that it matches package-lock.json**

🟠 **high · problem** · _effort: small_

**What:** `ensure_node_modules()` short-circuits with `return 0` whenever `node_modules/@playwright/test` is present in the persistent `ork3-fv-node-modules` volume (docker/entrypoint.sh:13), regardless of whether `package.json`/`package-lock.json` changed. Because that volume survives every container recreation (and, per finding #1, is shared across worktrees), switching to a branch that bumps Playwright keeps the old install forever with no re-install and no warning — while `/ms-playwright` holds the newer in-image browser build. That version/revision mismatch either hard-fails or resolves to a browser other than the recorded pin.

**Do:** Stamp the `package-lock.json` hash into the volume on install (e.g. `node_modules/.fv-lock-hash`) and re-run `npm ci` when it differs, even if `@playwright/test` is present. Alternatively include the lockfile hash in the volume name so a lockfile bump gets a fresh volume.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
```diff
@@ -22,31 +28,149 @@ Commands:
...
+    if [[ "$need_build" -eq 1 ]]; then
+      FUZZY_VALIDATOR_DOCKER_NETWORK="$net" \
+        docker compose -f "$COMPOSE_FILE" up -d --build
+    else
+      FUZZY_VALIDATOR_DOCKER_NETWORK="$net" \
+        docker compose -f "$COMPOSE_FILE" up -d
+    fi
+  )
+
+  local i
```
baltinerdist, line 136: **5. Readiness probe returns before the entrypoint's npm ci and socat forwards finish, racing the first real command**

🔴 **critical · problem** · _effort: small_

**What:** The wait loop polls `docker exec "$RUNNER_NAME" true` (bin/fuzzy-validator:137-142) and considers the runner ready the moment that succeeds. But `docker exec` spawns a sibling process in the container's namespace — it does not wait on PID 1. The entrypoint (PID 1) synchronously runs `ensure_node_modules` (a multi-minute `npm ci` on a fresh `fv-node-modules` volume) and then `start_db_port_forwards` (the socat listeners on 127.0.0.1:19306/19307 that the dual-profile ork-db wiring depends on) *before* backgrounding `sleep infinity` (entrypoint.sh:47-58). So the probe succeeds within a second or two and the wrapper immediately execs `fuzzy-validator-native` into a container whose `node_modules` is half-populated and whose DB ports are not yet listening. That is a genuine concurrent-write race on `npm ci` plus connection-refused on the DB ports, on exactly the first command a new contributor runs — surfacing as a raw module-resolution or DB error with no hint that it is a startup race. There is no compose healthcheck and no ready marker.

**Do:** Have the entrypoint remove `/tmp/fv-ready` at the top of each run and `touch` it only after `ensure_node_modules` and `start_db_port_forwards` both complete; change the wait loop to `docker exec "$RUNNER_NAME" test -f /tmp/fv-ready`. Raise the timeout for the cold-volume case, and after ~5s print "first run: installing node_modules, this can take several minutes" so the wait is legible.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
```diff
@@ -22,31 +28,149 @@ Commands:
...
+if [[ "${FUZZY_VALIDATOR_IN_CONTAINER:-}" == "1" ]]; then
+  exec "$NATIVE_BIN" "$@"
+fi
+
+want_native=0
+if [[ "${FUZZY_VALIDATOR_NATIVE:-}" == "1" ]]; then
+  want_native=1
+fi
+
+forward_args=()
```
baltinerdist, line 58: **6. Empty-array expansion under `set -u` aborts the script on stock macOS bash 3.2**

🟠 **high · problem** · _effort: trivial_

**What:** The script runs `set -euo pipefail` and execs `"$NATIVE_BIN" "${forward_args[@]}"` at line 69 — before the empty-command usage check at lines 73-77. Invoked as `bin/fuzzy-validator --host` with no subcommand, or `FUZZY_VALIDATOR_NATIVE=1 bin/fuzzy-validator` with zero args, `forward_args` is a genuinely empty array. On bash < 4.4 — exactly what ships as `/bin/bash` on macOS (3.2.57) and what `#!/usr/bin/env bash` resolves to without a Homebrew bash earlier on PATH — that throws `forward_args[@]: unbound variable` and aborts before any usage text. macOS plus the `--host` escape hatch is a documented workflow, so this is a realistic first-touch failure.

**Do:** Move the "no command given → usage" check above the native-mode exec, and/or guard the expansion for old bash: `exec "$NATIVE_BIN" "${forward_args[@]+\"${forward_args[@]}\"}"`. Optionally add an explicit bash-version guard since the script assumes bash-array semantics throughout.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
```diff
@@ -22,31 +28,149 @@ Commands:
...
+fi
+
+# Top-level help without docker: avoid forcing a build for bare --help.
+cmd="${forward_args[0]:-}"
+if [[ -z "$cmd" || "$cmd" == "-h" || "$cmd" == "--help" || "$cmd" == "help" ]]; then
+  usage
+  exit 0
+fi
+
+detect_php8_network() {
```
baltinerdist, line 79: **7. Docker missing or daemon down is reported as "php8 stack not running", and no failure path mentions the escape hatch**

🟠 **high · problem** · _effort: small_

**What:** `detect_php8_network()` treats any non-zero `docker inspect ork3-php8-app` as "ork3-php8-app is not running" and tells the developer to bring up the php8 stack (bin/fuzzy-validator:84-88). If Docker Desktop is not installed, not started, or the context is wrong, that advice fails too — sending a first-time user down the wrong path. Worse, `set -euo pipefail` means a missing `docker` binary aborts before any message prints at all. Separately, none of the three failure exits (network undetected, stack down, readiness timeout) mentions `--host` / `FUZZY_VALIDATOR_NATIVE=1`. An existing contributor whose native setup worked yesterday now hits a Docker error with no visible way back to what worked.

**Do:** Before `detect_php8_network`, check `command -v docker` and `docker info >/dev/null 2>&1` separately with distinct messages ("docker CLI not found" / "Docker daemon not reachable — start Docker Desktop"); only when the daemon is healthy and `ork3-php8-app` is genuinely absent should the message say to bring up the php8 stack. Append one line to every failure exit: "To run on the host instead (debug only, not the sign-off path): bin/fuzzy-validator --host <cmd>".

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
baltinerdist, line 136: **8. 90-second blind retry loop neither fails fast nor surfaces why the runner never became ready**

🟡 **medium · problem** · _effort: trivial_

**What:** `ensure_node_modules` can `return 1` (e.g. missing package-lock.json), which under the entrypoint's `set -e` kills PID 1 and exits the container almost instantly. But the wait loop only polls `docker exec … true`, sleeping 1s across all 90 attempts regardless of container state, then prints a generic "runner did not become ready". A fast, fully diagnosable entrypoint failure is presented as a slow, opaque 90-second hang with no exit code and no logs.

**Do:** Inside the loop also check `docker inspect -f '{{.State.Status}}' "$RUNNER_NAME"` and break immediately on `exited`/`dead`; before returning failure, print `docker logs --tail 50 "$RUNNER_NAME"`.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/docker/entrypoint.sh
```diff
@@ -0,0 +1,59 @@
...
+
+ensure_node_modules
+start_db_port_forwards
+
+# Prefer image-pinned Chromium over any host cache path.
+export PLAYWRIGHT_BROWSERS_PATH="${PLAYWRIGHT_BROWSERS_PATH:-/ms-playwright}"
+export FUZZY_VALIDATOR_IN_CONTAINER=1
+export ORK3_E2E_BASE_URL="${ORK3_E2E_BASE_URL:-http://ork3-php8-app/orkui/}"
+
+# Do not `exec` — keep socat children alive under this PID 1 shell.
```
baltinerdist, line 55: **9. Entrypoint's background+wait pattern never forwards SIGTERM, so every `docker stop` burns the full grace period**

💡 **low · enhancement** · _effort: trivial_

**What:** The entrypoint deliberately avoids `exec "$@"` (comment: "keep socat children alive under this PID 1 shell") and instead does `"$@" & child=$!; wait "$child"` with no `trap`. The stated rationale does not hold — `exec` replaces only the calling process's image; already-forked socat children are unaffected. As written, PID 1 is a bash script that ignores SIGTERM and never propagates it, so `docker stop`/`compose stop` on this long-lived runner always waits out the grace period before SIGKILL.

**Do:** Add `trap 'kill -TERM "$child" 2>/dev/null' TERM INT` before the `wait`, or restructure to `exec "$@"` after backgrounding socat — which preserves the socat children the comment cares about while giving proper signal delivery and exit-code semantics.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/docker-compose.runner.yml
```diff
@@ -0,0 +1,41 @@
...
+services:
+  fuzzy-validator-runner:
+    build:
+      context: ../..
+      dockerfile: tools/fuzzy-validator/docker/Dockerfile
+    image: ork3-fuzzy-validator-runner:local
+    container_name: ork3-fuzzy-validator-runner
+    restart: "no"
+    working_dir: /ork3
+    volumes:
```
baltinerdist, line 18: **10. Mounting /var/run/docker.sock grants the runner unrestricted host-root for a narrow use case**

🔴 **critical · problem** · _effort: large_

**What:** `docker-compose.runner.yml:21` bind-mounts the host Docker socket, sets `DOCKER_HOST=unix:///var/run/docker.sock`, and the Dockerfile installs `docker.io` + `docker-compose-v2` so the container drives the host engine directly. The justification is narrow (keeping the dual-profile `ork-db` workflow working, only exercised on `--ensure-sandbox`) but the grant is total: anything running inside this long-lived container — Playwright/Chromium rendering pages under test, `npm ci` pulling third-party packages at build *and* at container start, the Python compare pipeline — can create a privileged container, bind-mount host `/`, and read or exfiltrate anything on the developer's machine. Any RCE in the browser, an npm/pip dependency, or the tool's own code becomes full host compromise.

**Do:** Preferred: drop the docker.sock mount and have the host-side wrapper run the docker-touching steps (`bin/ork-db deploy-sandbox` and friends) on the host before/after `docker exec`-ing into the runner for browser and Python work. If in-container Docker control is genuinely required, front the socket with a scoped proxy (e.g. tecnativa/docker-socket-proxy) allow-listing only the specific inspect/exec calls `ork-db` needs against named containers, instead of the full Engine API.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/docker/Dockerfile
```diff
@@ -0,0 +1,68 @@
+# Fuzzy Validator 2.1 runner — Ubuntu 26.04 + Playwright-pinned Chromium.
```
baltinerdist, line 1: **11. No USER directive — the runner runs as root, and on Linux hosts every artifact it writes lands root-owned**

🟠 **high · problem** · _effort: medium_

**What:** The Dockerfile has no `USER` instruction, the compose service sets no `user:`, and `docker exec` is invoked without `--user` (bin/fuzzy-validator:175). Two consequences. Security: `npm ci`, the pip venv, Chromium, and the exec'd CLI all run as root, so combined with the docker.sock mount (#10) an attacker with code execution in any of them has root-on-host with zero extra steps. Ergonomics: with the whole repo bind-mounted read-write at `/ork3`, everything the runner writes on a Linux workstation — `reports/run-*`, `baselines/`, `calibrations/`, regenerated `manifests/<profile>/*.fuzz.json`, `setpoints/out/*.zip`, `.ork3-db.local` — is root-owned. `git clean`, `rm -rf reports/`, editor saves, and the `--host` escape hatch then all fail, pushing people to `sudo rm -rf` inside their worktree. Docker Desktop's ownership remapping hides this on macOS, which is why the arm64-only FV21-5 parity evidence never surfaced it, and none of the three docs mention it.

**Do:** Create a non-root user in the image (fixed UID, writable HOME, readable `/ms-playwright`) and run the steady-state process as it, and/or pass `--user "$(id -u):$(id -g)"` on the `docker exec` when the host is Linux. Have the entrypoint chown the `fv-node-modules` volume accordingly. If that is more surgery than this PR wants, at minimum document the Linux ownership consequence in the README's "ORK3 gotchas" and chown the tool's output dirs back to the invoking uid after each exec.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
```diff
@@ -22,31 +28,149 @@ Commands:
...
+
+ensure_runner_running || exit 1
+
+exec_env=(
+  -e FUZZY_VALIDATOR_IN_CONTAINER=1
+  -e "ORK3_E2E_BASE_URL=${ORK3_E2E_BASE_URL:-http://ork3-php8-app/orkui/}"
+  -e PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
+)
+
+# Pass through auth, clock, CI, and fuzzy/fuzz knobs when set on the host.
```
baltinerdist, line 155: **12. ORK3_E2E_PASSWORD is forwarded as a literal `docker exec -e KEY=VALUE` argument, visible in host process listings**

🟡 **medium · problem** · _effort: small_

**What:** The `pass_keys` loop forwards `ORK3_E2E_USERNAME`/`ORK3_E2E_PASSWORD` (the mirror profile's admin-like credentials) into `exec_env` as `-e KEY=VALUE` entries, which become literal argv on the `docker` client process (bin/fuzzy-validator:155-176). Anyone able to run `ps auxww`, read `/proc/<pid>/cmdline`, inspect Docker audit logs, or replay a recorded shell session sees the plaintext password for the duration of every command. This is new exposure: previously the value was inherited in-process with no cross-process argv hop.

**Do:** Stop passing secrets as `-e` argv. Write the needed vars to a per-invocation env file that the runner reads (bind a tmpfs path, or `docker exec` with stdin-fed env sourced by a small shim) so secrets never appear on the docker client's command line.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/docker-compose.runner.yml
```diff
@@ -0,0 +1,41 @@
+# Fuzzy Validator 2.1 — long-lived capture/validate runner (not part of php8 stack).
+#
+# Network: joins the external php8 user network. Default name matches compose
+# project "ork3"; override with FUZZY_VALIDATOR_DOCKER_NETWORK or let the CLI
+# wrapper auto-detect from `ork3-php8-app`.
+#
+# Lifecycle: restart "no" — auto-started by CLI; left running; does not survive reboot.
+
+services:
```
baltinerdist, line 9: **13. No defense-in-depth hardening on a container that necessarily holds elevated access**

💡 **medium · enhancement** · _effort: small_

**What:** Given the runner already holds the docker socket, runs as root, and has the full repo mounted read-write, the compose service adds none of the cheap standard mitigations: no `cap_drop`, no `no-new-privileges`, no resource limits. These do not close the docker.sock hole but materially raise the bar for a compromised dependency or a Chromium sandbox escape before it reaches the socket.

**Do:** Add `cap_drop: [ALL]` (re-adding only what Chromium's sandbox needs, or run Chromium with `--no-sandbox` inside the container as most CI setups do) and `security_opt: ["no-new-privileges:true"]` to the `fuzzy-validator-runner` service, plus modest cpu/memory limits.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/docker-compose.runner.yml
baltinerdist, line 18: **14. Repo bind-mount is fully read-write when only a handful of subpaths need to be writable**

💡 **medium · enhancement** · _effort: medium_

**What:** `- ../..:/ork3` mounts the entire repository read-write into a root container that executes third-party npm/pip code and browser automation. The tool's writable outputs are narrow and already enumerated in the README's Outputs section (baselines/, calibrations/, reports/, setpoints/out|cache/, overlays/putative, manifests/<profile>/*.fuzz.json); application source, `.git`, other tools, and CI config never need write access from inside this container.

**Do:** Mount the repo read-only (`../..:/ork3:ro`) and layer separate read-write binds or named volumes for just the tool's writable subpaths. This bounds what a compromised dependency or renderer can rewrite on the host to the tool's own scratch space, independent of #10 and #11.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/README.md
```diff
@@ -752,22 +752,36 @@ ORK3 uses this tool to gate front-end refactors against local docker. Two profil
```
baltinerdist, line 752: **15. Docs explain the network and npm-ci trade-offs in detail but never state the docker.sock/root trust boundary**

💡 **low · enhancement** · _effort: trivial_

**What:** tools/fuzzy-validator/README.md and USER-GUIDE.md both present the containerized runner as the new default sign-off path and carefully explain the node_modules-volume and `--ignore-scripts` mitigations, but neither mentions that the runner holds the host Docker socket and runs as root — a materially different trust posture from the host-native path it replaces. Developers defaulted into the container will not know it carries host-level blast radius unless they read the compose file.

**Do:** Add a short "Security model" callout to the README and USER-GUIDE quick-start sections naming the docker.sock mount, its purpose, and its risk, so future changes that widen what runs inside the container are evaluated against that trade-off.

<sub>Anchored to the nearest line inside this PR's diff; the finding refers to `tools/fuzzy-validator/README.md:749-816; docs/megiddo/fuzzy-validator/USER-GUIDE.md:23-51`.</sub>

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

docs/megiddo/fuzzy-validator/USER-GUIDE.md
```diff
@@ -22,20 +22,23 @@ It produces **exit code pass/fail** (for CI) and an **HTML report** (for humans)
...
 # Repo root
-npm ci
-npx playwright install chromium
-pip install -r tools/fuzzy-validator/python/requirements.txt
-
 docker compose -f docker-compose.php8.yml up -d
-bin/ork-db deploy-sandbox              # required for test profile
+bin/ork-db deploy-sandbox              # required for test profile (run on host)
 
+# Optional: host browsers / humans still use localhost
```
baltinerdist, line 32: **16. USER-GUIDE's setup block still exports the one env var that breaks the containerized default**

🔴 **critical · problem** · _effort: small_

**What:** USER-GUIDE.md:32-33 (touched by this PR — the old host-toolchain lines were removed but this export was kept and relabelled "Optional: host browsers / humans still use localhost") instructs `export ORK3_E2E_BASE_URL=http://localhost:19080/orkui/` in the same shell that then runs `bin/fuzzy-validator`. The wrapper forwards that value straight through: `-e "ORK3_E2E_BASE_URL=${ORK3_E2E_BASE_URL:-http://ork3-php8-app/orkui/}"` (bin/fuzzy-validator:151). Inside the runner, localhost:19080 is the runner itself — the PR's own FV21-0 spike proved this fails with curl exit 7 — so every capture fails to connect. Because people put these exports in a shell profile the failure is sticky and reads as a harness bug. The tool README got this right and deleted the export, so the two docs now actively contradict each other. The same block exports `ORK3_E2E_TEST_PASSWORD=test-db-player`, which is absent from the `pass_keys` allowlist (lines 156-168), so a non-default test password is silently dropped and auth falls back to the baked default.

**Do:** Remove the `ORK3_E2E_BASE_URL` export from the USER-GUIDE setup block (move it under the `--host` section where it is correct) and note that the runner supplies its own base URL. In the wrapper, detect a `localhost`/`127.0.0.1` host in `ORK3_E2E_BASE_URL` before exec'ing into the runner and either rewrite it to `http://ork3-php8-app/orkui/` with a one-line stderr note or refuse with an actionable message. Add `ORK3_E2E_TEST_PASSWORD` and any other documented `*Env` names from profiles.json5 to `pass_keys`.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
```diff
@@ -22,31 +28,149 @@ Commands:
...
+  FUZZ_TOOL_ROOT
+  FUZZ_PAGES_MANIFEST
+)
+for key in "${pass_keys[@]}"; do
+  if [[ -n "${!key:-}" ]]; then
+    exec_env+=(-e "${key}=${!key}")
+  fi
+done
+
+exec docker exec -i -w /ork3 "${exec_env[@]}" "$RUNNER_NAME" \
```
baltinerdist, line 175: **17. Host paths outside the repo, and the host working directory, silently stop working under the container default**

🟠 **high · problem** · _effort: small_

**What:** The final exec pins `-w /ork3` and mounts only the repo, node_modules, and docker.sock (bin/fuzzy-validator:175). Two new behaviors, neither documented. (1) Documented examples using paths outside the repo now resolve inside the container where `/tmp` is empty and `$HOME` is root's: `record --urls /tmp/pages.txt` (README.md:285), `validate --annotations /tmp/annotations.json` (README.md:343), `setpoint capture --out-dir /tmp/fuzzy-bundles` (README.md:477), and `setpoint restore --bundle ~/Downloads/….zip` (USER-GUIDE.md:211 — the exact flow after a Google Drive download). The developer gets "no such file" for a path they can `ls`. (2) The host cwd is discarded, so relative paths are now always resolved from the repo root and the same relative argument behaves differently in container vs `--host` mode.

**Do:** Detect arguments to `--urls`/`--annotations`/`--bundle`/`--out-dir` that resolve outside `$REPO_ROOT` and fail fast with "path X is outside the repo and is not visible inside the runner — copy it under the repo, or use --host"; optionally bind a scratch mount (`${TMPDIR:-/tmp}:/host-tmp:ro`, or `$HOME/Downloads`). Update the affected README/USER-GUIDE examples to repo-relative paths and state explicitly that container mode always runs with cwd = repo root.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

docs/megiddo/fuzzy-validator/README.md
```diff
@@ -4,7 +4,7 @@ Automated refactor-stability harness for Megiddo R-*: **hard CSS/JS byte checks*
 
 **Run:** `bin/fuzzy-validator record|validate|refuzz|setpoint …`  
 **Code:** `tools/fuzzy-validator/`  
-**Status:** v1 shipped (FU-0 … FU-16). Next product work: [version-2/](./version-2/).
+**Status:** v1 shipped (FU-0 … FU-16). v2 overlays shipped: [version-2/](./version-2/). **v2.1 runner shipped:** [version-2.1/](./version-2.1/) — default containerized Ubuntu 26.04 + pinned Chromium.
```
baltinerdist, line 7: **18. Docs hub says 2.1 is both "shipped" and "plan", still lists host Node/Playwright/Python as prerequisites, and never states the first-run cost**

🟡 **medium · problem** · _effort: small_

**What:** This PR flipped `docs/megiddo/fuzzy-validator/README.md:7` to "**v2.1 runner shipped**" but left line 44's folder table describing version-2.1/ as "Containerized runner plan … **status: plan**", and left the Prerequisites table (lines 68-77) listing "Node + Playwright — Root package.json" and "Python 3.11+ — pip install -r …/requirements.txt" as requirements — exactly what the default path removes and what USER-GUIDE.md:25 explicitly denies. This is the file "Start here" points new contributors at, so they install a host toolchain they will never use and are never told the two things that actually gate their first fifteen minutes: Docker must be installed and running, and the first invocation builds a multi-GB image (Ubuntu 26.04 + apt with node/npm/php8.5/docker.io, a pip venv, `npm ci`, `playwright install --with-deps chromium`) behind a single "starting runner…" line followed by long silence.

**Do:** Fix the shipped/plan contradiction. Rewrite the hub Prerequisites table for the default path (Docker running, ~N GB free disk, php8 stack, `bin/ork-db deploy-sandbox`, `setpoint restore`) and move Node/Playwright/pip into a clearly labelled "native escape hatch only" row. Add the `--host` / `FUZZY_VALIDATOR_NATIVE=1` hatch to the hub Quick start, plus a short "What changed in 2.1" banner telling existing native users their default execution path moved into a container and that gold setpoints must be re-recorded from the runner. In `ensure_runner_running`, print a one-time "first run builds the runner image — expect ~N minutes and ~X GB" banner before the initial `up -d --build`.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/docker-compose.runner.yml
baltinerdist, line 9: **19. No platform pin — macOS-arm64 and Linux-amd64 developers build different Chromium binaries from the same Dockerfile**

🟠 **high · problem** · _effort: medium_

**What:** Neither the compose file nor the Dockerfile sets `platform:`/`--platform`, so the image builds for the host architecture: linux/arm64 on Apple Silicon, linux/amd64 on a Linux x86 workstation. Playwright's arm64 and amd64 Chromium are separate builds with different SIMD raster paths and can rasterize text differently — precisely the drift this PR exists to eliminate, only now invisible because both developers are "in the runner". The team's own design doc already decided this (version-2.1/02-design.md:116: "linux/amd64 primary; document arm64 (Apple Silicon) build as required for local Mac") but nothing in the shipped files encodes it, `BROWSER_PIN.txt` records no architecture, and the FV21-5 parity evidence was gathered on a single macOS arm64 host — so acceptance criterion R1/AC1 ("same pass/fail as a second host using the same repo state") is untested across the two architectures the team actually uses.

**Do:** Pick one canonical arch for gold setpoints and make it explicit: either set `platform: linux/amd64` on the runner service (documenting the emulation cost on Macs in the README quick start), or keep native builds but add `arch=` to the generated BROWSER_PIN, stamp it into captures, and have `validate` refuse — not soft-warn — when the setpoint's arch differs from the running one. Either way, run the FV21-5 parity check once between an arm64 and an amd64 runner before calling AC1 met.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/docker/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+# node_modules strategy:
+#   Named volume `fv-node-modules` overlays /ork3/node_modules so Linux
+#   native addons never clash with a macOS host tree. Entrypoint runs
+#   `npm ci` into that volume when @playwright/test is missing.
+#   Playwright browser binaries live in-image at PLAYWRIGHT_BROWSERS_PATH
+#   (not the host ~/Library/Caches/ms-playwright).
+#
+# Pin: ubuntu:resolute-20260707 (== ubuntu:26.04 at spike time).
+
+ARG UBUNTU_IMAGE=ubuntu:resolute-20260707
```
baltinerdist, line 12: **20. Base image pinned to a mutable tag, and font/fontconfig versions float freely**

🟠 **high · problem** · _effort: small_

**What:** `ARG UBUNTU_IMAGE=ubuntu:resolute-20260707` (Dockerfile:12) is a tag, not a `sha256:` digest — it can be silently re-pushed and every subsequent `up -d --build` would pull different bytes under the same name. On top of that, every apt install (`fonts-liberation`, `fonts-dejavu-core`, `fontconfig`, plus whatever `playwright install --with-deps` pulls) floats to whatever the archive holds at build time. A freetype or font-package bump changes glyph metrics on every captured page — exactly the class of drift this harness gates on — and `BROWSER_PIN.txt` records none of those versions. This undermines the PR's own reproducibility goal, and it is a supply-chain gap as well as a determinism one.

**Do:** Pin `FROM` by digest (`ubuntu:resolute-20260707@sha256:<digest>`, keeping the readable tag in a comment — the FV21-0 spike already recorded a digest). Have the Dockerfile's pin generator append `dpkg-query` versions for fontconfig, freetype, and the font packages into the generated BROWSER_PIN, and fail the build if the generated pin differs from the committed `docker/BROWSER_PIN.txt`.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
```diff
@@ -22,31 +28,149 @@ Commands:
...
+  fi
+  forward_args+=("$arg")
+done
+
+if [[ "$want_native" -eq 1 ]]; then
+  echo "fuzzy-validator: native mode — captures may differ from containerized default" >&2
+  exec "$NATIVE_BIN" "${forward_args[@]}"
+fi
+
+# Top-level help without docker: avoid forcing a build for bare --help.
```
baltinerdist, line 72: **22. Subcommand `--help` now requires Docker and a running php8 stack**

🟡 **medium · problem** · _effort: trivial_

**What:** The help short-circuit at bin/fuzzy-validator:72-77 only catches the *bare* invocation, so `bin/fuzzy-validator validate --help` — advertised in the wrapper's own usage text (lines 43-44) and used as the rebuild incantation in tools/fuzzy-validator/README.md:775 — falls through to `ensure_runner_running` and dies with "ork3-php8-app is not running" whenever the stack is down. Reading flag documentation should never require bringing up two Docker stacks.

**Do:** If any element of `forward_args` is `-h`/`--help`, route straight to `"$NATIVE_BIN"` (or print the native usage) without touching Docker, and pick a different incantation for the README's rebuild recipe.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/bin/fuzzy-validator
baltinerdist, line 109: **23. Add `runner status|start|stop|rebuild|reset|shell` and `doctor` subcommands — today recovery requires knowing internal compose and volume names**

💡 **high · enhancement** · _effort: medium_

**What:** Every precondition (docker present, daemon up, php8 stack up, network resolvable, image built, node_modules populated, app reachable at http://ork3-php8-app/orkui/) is checked implicitly and late, surfacing as one terse line mid-run. Recovery is worse: the only documented reset is a raw `docker compose -f tools/fuzzy-validator/docker-compose.runner.yml stop` that omits `FUZZY_VALIDATOR_DOCKER_NETWORK` — the very value the wrapper exists to auto-detect, so it breaks on any non-default compose project — followed by `FUZZY_VALIDATOR_REBUILD=1`, whose ordering is load-bearing but unexplained (see #3). Nothing documents that a corrupt Linux node_modules tree needs `docker volume rm ork3-fv-node-modules`, a name that appears only in the compose file.

**Do:** Add `bin/fuzzy-validator runner <status|start|stop|rebuild|reset|shell>` that wraps compose with the detected network and reports the container's BROWSER_PIN, where `reset` = stop + `docker volume rm ork3-fv-node-modules` + rebuild; route the docs' manual compose snippets through it. Add `bin/fuzzy-validator doctor` printing a pass/fail line per precondition with the exact fix command for each failure.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

tools/fuzzy-validator/README.md
```diff
@@ -752,22 +752,36 @@ ORK3 uses this tool to gate front-end refactors against local docker. Two profil
...
 export ORK3_E2E_USERNAME=admin ORK3_E2E_PASSWORD=password
 # test profile defaults: megiddo / test-db-player
 
 bin/fuzzy-validator setpoint restore
 bin/fuzzy-validator validate --pages home-authenticated,player-profile --phase all
 ```
 
+First invocation builds/starts `ork3-fuzzy-validator-runner` if needed and **leaves it running** (`restart: "no"` — no reboot persistence).
+
+```bash
```
baltinerdist, line 772: **24. No documented teardown, and rebuilds silently accumulate dangling image layers**

💡 **medium · enhancement** · _effort: trivial_

**What:** The README documents `stop` and `--build` but no `down`/`down -v`/prune anywhere in README.md, USER-GUIDE.md, or the compose comments. Because the image is always built to the same `ork3-fuzzy-validator-runner:local` tag, every rebuild orphans the prior layer set as dangling, and `ork3-fv-node-modules` is never reclaimed. For a container the developer never explicitly provisioned — it auto-starts and stays up — this quietly consumes tens of GB on a laptop over months.

**Do:** Document a teardown recipe (`docker compose -f tools/fuzzy-validator/docker-compose.runner.yml down -v --rmi local`) in the README's rebuild/troubleshooting section, and expose it as `bin/fuzzy-validator runner reset|teardown` alongside #23.

---
<sub>Review performed by Claude, take with a grain of salt. -TofH</sub>

kenwalker:
Merged into master on 2026-08-18 as part of the 492+493+494 stack (this PR's head c558cf3f... is reachable from master). Closing to reconcile status — the containerized runner is live and verified (record+validate PASS on macOS, exit 0). Runner-file conflicts were resolved in favor of the newer 492-line versions.

PR #495, opened 2026-07-26
Enhancement: Walker release mobile support

baltinerdist:
Makes the Reeve & Corpora Qualification Tests module (Walker) usable on a phone.

A mobile review of the module measured **45 distinct defects** across the take flow, the admin surfaces and the qualified-player reports, at 320×568, 390×844, 768×1024 and 844×390. All 45 are addressed here.

Before this branch the module had **two** media queries across ~7,900 lines of template — both `600px`, covering only the take flow and the profile cards. `QualTest_manage`, `QualTest_questions`, `QualTest_question` and both `Reports_*` templates had none at all.

## Root causes

Most of the 45 traced back to a handful of causes rather than 45 separate bugs:

- **`.rp-table-area` / `.rp-stat-card` were content-box with no `box-sizing`.** `max-width:100%` clamped the *content* box, so every `rp-*` page rendered its card 34px wider than its parent and 27px was clipped where `html`/`body` overflow made it unreachable. On the questions list this left the report-question control 17 of its 20px off-screen at *every* scroll position.
- **Configure Tests hardcoded an inline `grid-template-columns:1fr 1fr`.** Being inline, no media query could reach it; the tracks resolved to ~205px each regardless of viewport, putting the entire Corpora card and its sharing toggle off-screen. Now a `.qt-config-grid` class that collapses to one column at 768px.
- **Every module modal combined a desktop `min-width` floor, content-box sizing and a zero-padding overlay**, so five of them rendered wider than the viewport with no backdrop gutter. All are now `border-box`, `min-width:0`, with a 12px overlay gutter and internal scroll.
- **Neither the take flow nor the in-profile quiz did any scroll management**, so the question stem landed above the viewport after Next and the pass/fail verdict was scrolled off entirely.

## Also included

- Answer choices are a real radiogroup — `role`, `aria-checked`, `aria-setsize`/`aria-posinset`, roving `tabindex`, arrow-key handling
- **"Select all that apply" questions now show a square checkbox instead of a round radio** — multi-answer options carried the ARIA `checkbox` role but drew the same round radio as one-of-N questions, so nothing visually said more than one answer was allowed. They now render a square box with a tick, in both themes and all four states (unselected, selected, correct, incorrect). Single-answer questions keep the round radio. This one applies on desktop as well as mobile.
- Correct/incorrect feedback is announced via `role="status"` and carries an icon, rather than relying on colour alone
- Progress ("Question 3 of 5") exposed with `aria-live`
- Touch targets across the module reach 44px; inputs reach 16px so iOS does not zoom on focus
- Dark mode covers the version-contents and previous-versions panels, which were rendering as white slabs inside a dark modal
- `prefers-reduced-motion` guard on the result-view animations

## Scoping

**Desktop layout is unchanged.** Every layout fix is scoped inside a media query or a module class, except `box-sizing`/aria/focus/dark-mode corrections that were measured as no-ops at 1440px. The one deliberate exception is the multi-answer checkbox above, which is a correctness fix for the affordance itself and so applies at every width.

`reports.css` is shared by ~25 pages, so the content-before-sidebar stacking is **opt-in**: a `.qt-page` class on the five module pages that have a sidebar. A blanket `order` flip would have reordered every `rp-*` consumer, several of which put filter controls in the sidebar where sidebar-first is the intended reading order.

## Verification

Exercised live at 320 and 390 in both themes — built a question and saved it, ran the admin preview/bulk-import modals, and took a test end to end.

| Surface | Before | After |
|---|---|---|
| `.rp-table-area` @320 | W341 R347 (27px unreachable) | W308 R314 |
| Questions DataTable @320 | 686px | 419px, scrolls in-container |
| Report-question button | 20px wide, 17px off-screen | 44×44, reachable |
| Configure Tests @320 | Corpora card at x249–455, toggle off-screen | single column, toggle visible |
| Manage / take / question form @320 | overflow offenders | 0 offenders |
| Dark `.qt-vq`, `.qt-ver-past` | `#fff` | `rgb(55,65,81)` |

The only sub-44px targets remaining on module pages are app-wide chrome (global nav, theme toggle, footer), which are outside this module's scope.

https://claude.ai/code/session_01NK9CAyxMGAufyamzP4NCdJ


baltinerdist:

Synced `feature/walker-mobile-fixes` up to `master` @ `6ab028a3` (merge commit `4403cb24`). This branch was **263 commits behind**.

- **Conflicts:** none — merged clean.
- **Method:** `git merge upstream/master` rather than a rebase. This branch's history contains prior merge commits; a linear rebase would replay them against the new base and manufacture conflicts that don't really exist.
- **Verification:** no conflict markers anywhere in the tree; no unintended files in the merge.

PR now shows `MERGEABLE / CLEAN`. No changes to this branch's own commits — the feature work is untouched.

PR #497, opened 2026-08-09
Enhancement: ORK logo on home page masthead, four-widget stats bar

baltinerdist:
Adds the new ORK wordmark to the top-left of the home page stats bar and reduces the bar from five widgets to four, merging the separate Kingdoms and Principalities counts into one **Kingdoms & Principalities** figure.

`[ORK logo] │ 27 Kingdoms & Principalities │ 331 Parks │ ~2,605 Players/Week │ Weekly Recap →`

## Changes

**New assets** — `assets/images/logo-ork-wide.svg` and `logo-ork-wide-ondark.svg`, both pure vector with no embedded rasters.

They are referenced via `` rather than inlined on purpose: both files declare `id="Layer_2"` and an identical `.cls-1` … `.cls-14` class set inside a `<style>` block, so inlining them into the same document would cross-contaminate their fills. As separate documents they can't collide.

**The logo cell is the page `<h1>`.** The previous "Welcome to the Amtgard Online Record Keeper" heading is removed — the logo reads *O.R.K. — online record keeper*, so the two duplicated each other within ~80px. Wrapping the logo in `h1` with matching `alt` text keeps the document outline intact rather than leaving the page with no top-level heading. Removing that heading also orphaned the MedievalSharp Google Fonts link (its only consumer), so that render-blocking request is dropped too.

**Widgets are nested in a `.hm-stats-cells` wrapper.** This is structural, not cosmetic: the existing mobile rules do 2-column border math with `:nth-child(odd)` / `:nth-last-child(2)` on `.hm-stat-item`. Adding the logo as a *sibling* would offset every index by one and silently corrupt the grid. Nesting keeps all that math untouched.

**Dark mode** swaps variants on `html[data-theme="dark"]`, which is how the app toggles themes (`<picture>` with media queries would not respond to it). All three display toggles are scoped to `.hm-logo-cell` so they outrank the `.hm-logo-cell img` sizing rule — an unscoped `.hm-logo-dark { display: none }` loses on specificity and renders both logos side by side in light mode only.

**The combined count** is computed live as `count($hmKingdoms) + count($hmPrinz)`. The two arrays are populated by a mutually exclusive branch on `ParentKingdomId === 0`, so they're disjoint and the sum is exact.

## Sizing

Logo renders at 110px on both desktop and mobile (no mobile override — one value to maintain). `max-width: 100%` plus `object-fit: contain` scales it down proportionally on viewports narrower than ~252px instead of distorting it. The stats bar grows from ~60px to 127px on desktop.

## Verification

Checked against a local instance in both themes:

| Case | Result |
|---|---|
| Light mode | Light variant only |
| Dark mode | On-dark variant only |
| Desktop, 4 widgets | Four equal cells, no label wrap |
| Mobile 390px / 320px | Clean 2×2 grid, no horizontal overflow |
| `php -l` | No syntax errors |

The Weekly Recap cell is unchanged — its `!empty($LoggedIn) && is_array($week_recap ?? null)` conditional is byte-identical, differing only by one tab of indentation from moving inside the wrapper. It doesn't render on a local DB whose `ork_weekly_recap` table has no row for the current week, which is pre-existing behaviour; the 4-widget layout above was verified with the cell injected.


kenwalker:
Quick tested this including mobile views and dark/light modes, looks good. Once we get some stuff synced up this should be an easy one.

baltinerdist:

Synced `feature/new-logo-add` up to `master` @ `6ab028a3` (merge commit `2cb460ed`). This branch was **262 commits behind**.

- **Conflicts:** none — merged clean.
- **Method:** `git merge upstream/master` rather than a rebase. This branch's history contains prior merge commits; a linear rebase would replay them against the new base and manufacture conflicts that don't really exist.
- **Verification:** no conflict markers anywhere in the tree; no unintended files in the merge.

PR now shows `MERGEABLE / CLEAN`. No changes to this branch's own commits — the feature work is untouched.

PR #506, opened 2026-08-19
Bugfix: pass Token to one_shot and get_authorization_list

kenwalker:
Follow-up to #505. Two more callers of token-gated services that supply no `Token` — the same latent defect, found by auditing the rest of the codebase.

## Why these broke

Until `307dcad0` ("Polish: harden ork_session helpers"), `IsAuthorized_h()` began with:

```php
if (isset($_SESSION['is_authorized_mundane_id']))
    return $_SESSION['is_authorized_mundane_id'];   // before the token is even examined
```

Any call with a missing token still resolved to the logged-in user, so these call sites worked for years. That short-circuit was removed — correctly, since it accepted *any* token once a session existed — which turned them into live failures.

## 1. Admin face-image upload (`controller.Admin.php`)

`one_shot()` is a pass-through to `Player::AddOneShotFaceImage()`:

```php
$requester_id = Ork3::$Lib->authorization->IsAuthorized($request['Token']);
if (valid_id($requester_id) && HasAuthority($requester_id, AUTH_PARK, $mundane['ParkId'], AUTH_CREATE)
    || $requester_id == $request['MundaneId']) {
```

With no `Token`, `$requester_id` is `0`: `valid_id(0)` is false and `0 == $request['MundaneId']` is false. Both branches fail and the method falls out with no explicit return — the image is never sent to Behold, and nothing surfaces an error to the uploader.

## 2. Admin authorization list (`model.Reports.php`)

`get_authorization_list()` built its request with `Type`, `Id`, `Officers` and no `Token`. `Report::GetAuthorizations()` gates its result set on:

```php
&& ($mundane_id = Ork3::$Lib->authorization->IsAuthorized($request['Token'])) > 0
&& Ork3::$Lib->authorization->HasAuthority($mundane_id, $request['Type'], $request['Id'], AUTH_EDIT)) {
```

so the page received an empty list rather than an error. Fixed in the model, matching `dues_paid_list()` and `player_roster()` in the same file.

## Audit scope

Extracted the 143 service methods in `system/lib/ork3/` that resolve an actor via `IsAuthorized($request['Token'])`, mapped the 108 model methods forwarding to them, and checked every call site in `orkui/controller/` and `orkui/model/` for a token supplied either as a `'Token'` key or positionally. These were the only two missing one.

Caveat: the audit follows model pass-throughs one level deep, so a request assembled several frames up wouldn't be caught. High confidence, not exhaustive.

## Testing

Not verified end to end. Worth exercising after deploy: upload a player face image from the admin panel and confirm it appears, and load the admin authorization page and confirm the non-officer list is populated.

### DIFF
--- a/orkui/controller/controller.Admin.php
+++ b/orkui/controller/controller.Admin.php
@@ -978,6 +978,7 @@ class Controller_Admin extends Controller
                                     $face_im = file_get_contents(DIR_TMP . sprintf("fi_%06d", $id));
                                     $face_imdata = base64_encode($face_im);
                                     $one = $this->Player->one_shot([
+                                        'Token' => $this->session->token,
                                         'MundaneId' => $id,
                                         'Base64FaceImage' => $face_imdata
                                         ]);
--- a/orkui/model/model.Reports.php
+++ b/orkui/model/model.Reports.php
@@ -172,6 +172,7 @@ class Model_Reports extends Model
     public function get_authorization_list($type, $id, $officers)
     {
         $request = array(
+                'Token' => $this->session->token,
                 'Type' => $type,
                 'Id' => $id,
                 'Officers' => $officers
--- a/tests/Unit/TokenGatedCallSiteTest.php
+++ b/tests/Unit/TokenGatedCallSiteTest.php
@@ -1,0 +1,280 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Guards the controller/model -> service seam for missing Token arguments.
+ *
+ * Service methods in system/lib/ork3 resolve the acting user with
+ * IsAuthorized($request['Token']) and gate their writes on HasAuthority().
+ * A caller that omits Token therefore resolves to mundane_id 0 and is refused.
+ *
+ * Until 307dcad0 ("Polish: harden ork_session helpers") this was survivable:
+ * IsAuthorized_h() short-circuited on $_SESSION['is_authorized_mundane_id']
+ * before it ever inspected the token, so a missing Token still resolved to the
+ * logged-in user. Removing that short-circuit was correct -- it accepted any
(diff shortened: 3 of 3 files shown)


kenwalker:
Added `tests/Unit/TokenGatedCallSiteTest.php` (f752d6a1) — a static contract test guarding the seam that produced both this PR's bugs and #505's.

**Why the existing suite missed all three:** `phpunit.xml.dist` scopes coverage to `system/lib/ork3` and `orkservice`. `orkui/` is not included. The 90 existing tests exercise the service layer, where the auth logic is correct — `AddDues` properly refuses a request with no token. The defect is in the caller failing to supply one, and that wiring was untested.

**Two checks, matching the two shapes found in the wild:**
- **A** — a call site passing an array literal to a token-gated model method (`Controller_Admin` `adddues`, `one_shot`)
- **B** — a model method building a request literal for a token-gated service (`Model_Reports::get_authorization_list`)

A third test asserts the detector still matches known token-gated methods (`AddDues`, `AddAward`, `AddOneShotFaceImage`, `GetAuthorizations`), so renaming `IsAuthorized` or changing the request key can't make the other two pass vacuously.

**Verified by mutation** — removing the `Token` from each fixed call site fails the corresponding check, and only that one:

| Mutation | Check A | Check B |
|---|---|---|
| `adddues` token removed | **FAIL** | pass |
| `one_shot` token removed | **FAIL** | pass |
| `get_authorization_list` token removed | pass | **FAIL** |

**Scope limits, stated plainly:** it's a source heuristic. Positional and opaque-variable call sites are out of scope — not statically resolvable, and covered by service-level tests instead. Model pass-throughs are followed one level. `ALLOWED` is left empty as the documented place to record intentional exceptions.

**Note on verification:** there's no `vendor/` in this checkout and the app container runs PHP 8.1 (PHPUnit 11 needs 8.2+), so I exercised the test logic through a stub harness rather than PHPUnit itself. Worth a run under the real runner in CI before merge.

PR #509, opened 2026-08-20
Bugfix: Live dashboard silently empty — Clear() before DataSet on the shared $DB

kenwalker:
## Summary

The **/Live** attendance dashboard renders empty on production even when plenty of sign-ins qualify — at diagnosis time, **173 credits in the rolling 24h window across 25 parks/events** were confirmed by running the dashboard's own aggregate SQL directly against prod, while `Live/stats` returned `{"active_3h":0,"parks":[],"events":[]}`.

## Root cause

The seventh known occurrence of the **yapo bound-param hazard**: `YapoMysql::DataSet()` executes with whatever bindings are sitting in `$this->Data` —

```php
if (is_countable($this->Data) && count($this->Data) > 0) {
    $Query->execute($this->Data);   // stray params on a placeholder-free query
}
```

— and `class.Live` was the only domain code calling `DataSet()` on the **shared global $DB** without `Clear()`ing first. Session/nav machinery earlier in the request leaves bindings behind; PDO rejects the parameters on a query with no placeholders; under `ERRMODE_WARNING` the failure is silent, so every Live query returned an empty set. Identical mechanism to the `DestroySessionsForUser` fix that shipped with #491.

## Fix

`$this->db->Clear()` before all 7 `DataSet()` call sites in `class.Live.php`, with a one-line comment naming the hazard. Every other `DataSet` caller in the codebase already follows this convention.

**Diff note**: most of the churn is php-cs-fixer converting the file's tab-era indentation to house style; the semantic change is the 7 `Clear()` lines.

## Verified

- Reproduced locally: freshened 5 mirror rows' `entered_at`, logged-in hit of `Live/stats` → empty (matching prod). With the fix → `active_3h: 5` and fully hydrated park entries. Test rows restored afterward.
- `Live/recent` returns a clean empty shape when genuinely nothing is in-window (previously indistinguishable from the failure).
- 367/367 tests pass; `php -l` and fixer clean.

## Deploy

`git pull` only — PHP rides the bind mount, and the stats/ticker ghettocaches expire in ≤60s on their own.

## Follow-ups (deliberately not in this PR)

1. **Midweek under-reporting** (separate, pre-existing): the `date >= yesterday` index prefilter drops credits entered late for older game dates — only 25–46% of weekday-entered credits are visible to Live even when the query works. A superset floor (`date >= CURDATE() - INTERVAL 45 DAY`) fixes the counts while keeping the index useful — but it changes what "Live" means (parks would light up on data entry, not just on game day), so it deserves a design decision rather than a drive-by.
2. **Repo-wide audit / lint rule**: a quick scan suggests a few other `DataSet` calls may lack a nearby `Clear()` (`class.Player` shows 4 candidates). Seven occurrences of the same hazard argues for a CI grep or a `DataSet()` that resets `$this->Data` after execute.

### DIFF
--- a/system/lib/ork3/class.Live.php
+++ b/system/lib/ork3/class.Live.php
@@ -14,50 +14,52 @@
  * UTC. We compute cutoffs in PHP (Chicago TZ) and pass them as literal strings
  * so the comparison is apples-to-apples.
  */
-class Live extends Ork3 {
+class Live extends Ork3
+{
+    public const CACHE_TTL_STATS  = 60;   // seconds; matches the client poll interval
+    public const CACHE_TTL_RECENT = 20;   // ticker polls faster than the stats panel
+    public const RECENT_LIMIT     = 50;
 
-	const CACHE_TTL_STATS  = 60;   // seconds; matches the client poll interval
-	const CACHE_TTL_RECENT = 20;   // ticker polls faster than the stats panel
-	const RECENT_LIMIT     = 50;
+    public function __construct()
+    {
+        parent::__construct();
+    }
 
-	public function __construct() {
-		parent::__construct();
-	}
+    /**
+     * Returns:
+     *   {
+     *     now:        'YYYY-MM-DD HH:MM:SS',         // chicago wall-clock at query time
+     *     active_3h:  <int>,                          // total signins in past 3h
+     *     parks:      { <park_id>: { name, kingdom, city, province, title, lat, lng, day, h3, m30 } },
+     *     events:     { <event_id>: { name, kingdom, lat, lng, coord_source, day, h3, m30 } }
+     *   }
+     */
+    public function stats()
+    {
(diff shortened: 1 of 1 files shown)


kenwalker:
Added the regression guard to this PR: `DataSetClearLintTest` statically scans the domain layer and fails if any `DataSet()` call lacks a preceding `Clear()`/`SetData()` in its function. Proven both directions — green on this branch, reports exactly the 7 violations when pointed at master's `class.Live.php`.

Also closing the loop on the repo-wide audit promised in the PR body: **`class.Live` was the only offender.** All 290+ other `DataSet` sites across the 21 domain classes follow the Clear-first convention (the `class.Player` candidates from the crude scan have their `Clear()` a few lines further up than the grep window; the others share a `Clear()` across if/else branches). So this PR + the lint test fully retire the hazard class.

